PR #6, opened 2016-09-14
README for Github, with authors as per v1.7.0

jwuttke:


gpospelov:
Trying to reject since we agered on another README/README.md structure

PR #9, opened 2016-09-21
Provide way to include specular peak

waltervh:


gpospelov:
Great, we are getting closer! Glad to have you on board!

PR #14, opened 2016-09-28
revised progress handler updating, and restored cancellation

jwuttke:
ProgressHandler::incrementDone() now called more rarely than before: only once every 1000 steps. However, this has only minor, if any, impact on the total time taken by 'ctest'. Therefore I cannot confirm that the mutex in ProgressHandler::incrementDone() was slowing down the simulation.

Anyway, cancellation is restored, and manually tested under the GUI. Therefore I consider issue 1603 resolved.


jwuttke:
With 333338ad, remaining problems are all resolved:
* cancellation works without delay
* display of partial results is restored

PR #22, opened 2016-10-11
BornAgain-specific configuration for patched pdf2html

jwuttke:
used to generate http://apps.jcns.fz-juelich.de/src/BornAgain/BornAgainManual-1.5.1.html


waltervh:
Can we remove the test pdf, which does not really belong to the repository?

jwuttke:
ok, test.pdf is removed

PR #24, opened 2016-10-11
New procedure to prepare release news, and first draft for 1.7

jwuttke:
We should try to formulate our release more attractively, in more complete sentences, and in less technical terms. This cannot be done in the last minute of release day; the text needs some time to mature, and it needs consideration of all of us. Therefore I propose that from now on release news are prepared by successive edits to CHANGELOG.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,7 +9,7 @@ BornAgain-1.6.1, released 2016.07.15
 
 BornAgain-1.6.0, released 2016.06.30
   > Summary:
-    1) Python 3 support
+    1) Python3 support
     2) GUI: beta version of fitting
     3) Core: new roughness calculation that is more stable for large roughness
     4) Core: new formfactors dodecahedron and icosahedron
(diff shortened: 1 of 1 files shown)


gpospelov:
Please have in mind http://apps.jcns.fz-juelich.de/redmine/issues/1521, before it is solved, the very first line of CHANGELOG must have format exactly as "BornAgain-1.6.2, released 2016.07.15"

jwuttke:
What I am proposing here is meant for the development phase. Before merging into master, the standard header line has to be added. Therefore I don't anticipate any conflict with the GUI executable,
which looks up the 'master' version of CHANGELOG, right?

jwuttke:
Actually, bornagain retrieves CHANGELOG from http://apps.jcns.fz-juelich.de/src/BornAgain/CHANGELOG. Copying files to that location comes very late in the release procedure. So I do not see any conflict at all.

gpospelov:
Right. One of the step of the release procedure is to copy CHANGELOG from release-1.X.X branch (which will be a snapshot of the development branch) to apps server. There are 3 solutions possible:

1) solve issue #1521, when we are free in modifying CHANGELOG
2) Modify release procedure to add on the fly necessary line 
3) Make sure that current CHANGELOG in the development contains necessary line, to not to deal with it 3 weeks later

To my opinion, to keep things simple, we just have to make 3) now. This will solve all our problems for the next 2 months. Later, one day, we will solve 1). But I would not try to solve the problem via 2) this release. It will be just a temporary hack which anyway will require further involvement as soon as 1) is done.

gpospelov:
And may be I wasn't clear enough. Of course, for the moment you can do whatever you want with CHANGELOG, it want affect GUI update notifier. But at the release day the necessary line should be there. One way to not to forget this, would be to place line "BornAgain-1.6.99, released 20XX.XX.XX" already now.

jwuttke:
Before releasing the CHANGELOG, we have to paste the full list of resolved issues. Is that done manually or through a script? Can't we insert the header line at that occasion?

gpospelov:
So far it is done manually by copy and pasting issues from redmine page. We can discuss the release procedure today, if you want.

PR #25, opened 2016-10-12
interference functions: little simplifications, and improved comments

jwuttke:


waltervh:
Can we undo the renaming of FormFactorInfo? This class is meant to contain information that is needed for internal computations, but which does not belong in the form factors themselves. Although it currently only contains the relative abundance of the particle in question, it was wrapped to provide for possible extensions.

gpospelov:
Please consider the code in GUI/externals as a kind of legacy code. It's good to have it intact unless the bug is found. This simplifies comparison against the source (e.g. qt-creator source code) and making decision about possible updates. Could you please return GUI/external to previous state?

jwuttke:
Pull request closed since I continue to work on the interference function machinery.

jwuttke:
@gpospelov: yes, I should restore GUI/external before reopening this pull request.

jwuttke:
agreed with Walter: FormFactorInfo aka WeightedFormFactor becomes FormFactorWrapper.

jwuttke:
re-opened for review before it becomes too long; I'll now continue to work in a different branch.

PR #27, opened 2016-10-14
manual: improvements here and there, and revised chapter on Born approximation

jwuttke:
Improvements here and there, especially in the index and the list of symbols.

Revised chapter on Born approximation: Redefined index of refraction no longer depends on r, no longer contains local fluctuations. Dispersion relation is now derived, and generalized to n!=1. The Born approximation is now derived for generic K = K_0 n.


jwuttke:
I renamed \chi -> \delta v because it is not quite the electric susceptibility; with the old notation we would have run into conflict when writing the chapter on electromagnetic wave propagation.

PR #33, opened 2016-10-19
Region of interest in functional tests

gpospelov:
With this region of interest is fully implemented in the simulation.

* Fixed bug in axes range update in GUI's ColorMapPlot family
* ROI propagated from Core to GUI and back.
* ROI can be exported from Core to Python scripts.
* Masks are working in GUI/PyExport in the case of rectangular detector.
* 6 new functional tests

Next step will be to make fitting respect ROI.


gpospelov:
good idea

PR #37, opened 2016-10-21
Fitting with roi

gpospelov:
This was activity towards implementing ROI in fitting. Not yet done. Would like to merge to propagate changes to others.

* Move constructor in SimulationElements
* GISASSimulation getOutputData removed
* Detector now creates intensityDataMap out of vector elements without intermediate GISASImulation::m_intensity_data


Core/Binning/SimulationElement.cpp
```diff
@@ -39,6 +39,20 @@ SimulationElement::SimulationElement(const SimulationElement &other)
...
+    , m_intensity(std::move(other.m_intensity))
+    , m_polarization(std::move(other.m_polarization))
+    , m_analyzer_operator(std::move(other.m_analyzer_operator))
+    , mP_pixel_map(std::move(other.mP_pixel_map))
+    , m_contains_specular(std::move(other.m_contains_specular))
+{
+
+}
+
+
```
waltervh: Remove std::move on built in types; they don't have move constructors

Core/Instrument/SimulationArea.cpp
```diff
@@ -77,3 +77,17 @@ size_t SimulationArea::totalSize() const
...
+        return static_cast<int>(globalIndex);
+
+    size_t nxGlob = m_detector->getAxisBinIndex(globalIndex, BornAgain::X_AXIS_INDEX);
+    size_t nyGlob = m_detector->getAxisBinIndex(globalIndex, BornAgain::Y_AXIS_INDEX);
+    int nx = nxGlob - m_roi_x1;
+    int ny = nyGlob - m_roi_y1;
+    if(nx<0 || nx > static_cast<int>(m_roi_x2-m_roi_x1+1)) return -1;
+    if(ny<0 || ny > static_cast<int>(m_roi_y2-m_roi_y1+1)) return -1;
+    return static_cast<int>(ny + nx*(m_roi_y2-m_roi_y1+1));
+}
```
waltervh: If all internal indices are size_t, why not make this method return a size_t and remove the static_cast?

gpospelov: It can return -1, if globalIndex is outside of ROI area.

gpospelov: In principle you are right, I'm not checking it, and expect iteration always going inside. That's, after all, the SimulationArea is for. Ok.

jwuttke:
is there a peculiar reason to use "noexcept" with SimulationElement::SimulationElement(SimulationElement &&other) ?

or do you propose to slowly migrate towards using the "noexcept" and "throw" specifications everywhere ?

jwuttke:
What about this alternative solution for the SimulationElement constructors  :?

Inherit ICloneable, replace the copy constructor by a clone() function, and get the default move constructor for free.

gpospelov:
It's about vector.push_back(std::move(simulationElement)). Pushing itself is going via move, no matter whether 'noexcept' is present. But there is also underlying vector::resize, which is accomplished by vector. When there is no 'noexcept', this resize is going through copy constructor. With 'noexcept', vector::resize also using move.

jwuttke:
http://en.cppreference.com/w/cpp/container/vector/resize
says about
void std::vector::resize( size_type count )
the following:

If T's move constructor is not noexcept and T is not CopyInsertable into *this, vector will use the throwing move constructor.

This conflicts with your explanation that the copy constructor is used.

gpospelov:
About clone: simulationElement is low level type, even without virtual destructor. It is intended to be fast, as light as possible and shouldn't introduce inheritance. Also I don't see how clone() operation can replace move semantic.

About resize behavior: I wrote tests which demonstrates behavior I mentioned. One have to look into it more carefully.

jwuttke:
Inheritance from a base class as light and trivial as ICloneable won't affect performance, but improve consistence of our coding style. Deleting the copy constructor should enable the default move constructor, so the explicit implementation could be dropped.

As for the noexcept, which you found empirically to be necessary: why not add a brief comment to the code like "// noexcept needed for vector::resize" ?

gpospelov:
Inheritance adds some bytes because of virtual table, at least. It's also shouldn't be misused or overused. Anyway, suggest to move this discussion to other place. 

As for the move constructor, one can consider, after tests, switching to SimulationElement(SimulationELement&&) = default; This will do the thing, probably, without the need to delete copy constructor.

PR #44, opened 2016-10-27
Fit from gui

gpospelov:
Masks and ROI are propagated from ImportDataView to the GUI fit job. Now it should be possible to fit small areas of large detectors (with masks also), as in FitGALAXI example. This concludes, in general http://apps.jcns.fz-juelich.de/redmine/issues/1488. Lots of validation is required.


waltervh:
Try to remove code instead of commenting it out

gpospelov:
Right, will do it on the next iteration.

PR #50, opened 2016-11-01
Light refactoring of parameter and particle distribution machinery

jwuttke:
While working towards understanding of ParticleDistribution::generateParticles():
Comments, code layout, move small code fragments around.

### DIFF
--- a/Core/Aggregate/ParticleLayout.cpp
+++ b/Core/Aggregate/ParticleLayout.cpp
@@ -145,8 +145,8 @@ SafePointerVector<const IParticle> ParticleLayout::getParticles() const
     SafePointerVector<const IParticle> particle_vector;
     for (auto particle: m_particles) {
         if (const auto* p_part_distr = dynamic_cast<const ParticleDistribution*>(particle)) {
-            std::vector<const IParticle*> generated_particles;
-            p_part_distr->generateParticles(generated_particles);
+            std::vector<const IParticle*> generated_particles =
+                p_part_distr->generateParticles();
             for (const IParticle* particle: generated_particles)
                 particle_vector.push_back(particle);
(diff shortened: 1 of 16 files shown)


jwuttke:
nobody reviewing right now, I suppose? Please hold, one more commit to follow.

gpospelov:
We are few days before the release and I think it is a bad timing to make such kind of "light refactoring" just for the sake of understanding. The very few hours we have left should be devoted to the critical bug fixing. The rest, to my opinion, should wait until the next release. We had already very negative experience of last minute activity of such kind. I would like to avoid it and have stabilization period.

Also, to my opinion, such kind of refactoring should be supplemented with more active unit testing. Most of tests were written by practica student and certainly needs revision.

jwuttke:
There are two aspects to this: whether it's wise for me to do this kind of refactoring, and whether it's appropriate to merge right now.
- "For the sake of understanding" for me is a necessary step before starting the next section of the Manual.
- "We are few days before the release" is a valid objection to merging. So just leave this pull request on hold until after the release.

waltervh:
ParameterDistribution needs to be IParameterized instead of INamed. Some classes do not yet properly register their parameters for fitting or other purposes. In the end, we will need to move more classes to IParameterized and override addParametersToExternalPool() to correctly get the names of children's parameters (this is done more or less automatically for ICompositeSample classes).
Along the same lines, the name of ParameterDistribution cannot be the name of its main parameter, since it can contain '/'. Building up a tree of parameters will be buggy, or at the least very confusing, if single objects can have slash separated names. On the other hand, "std::string m_name;" for the main parameter name in ParameterDistribution was a bad choice of name. I suggest "main_par_name" or something similar here.

jwuttke:
Freshly starting from develop, I reenacted all changes (using git cherry-pick), except for those concerned with IParameterized->INamed.

PR #69, opened 2016-11-11
Add patched FindBoost.cmake, needed when CMake-3.6 hits Boost-1.62

jwuttke:
Otherwise build will fail under latest Debian/testing.


jwuttke:
for a long-term solution, I reported the problem at http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=843997

PR #83, opened 2016-11-24
resolved all "using namespace BornAgain"

jwuttke:


GUI/coregui/Models/TransformFromDomain.cpp
```diff
@@ -56,8 +56,6 @@
 #include <QDebug>
 #include <limits>
 
-using namespace BornAgain;
-
```
waltervh: Since cpp files are not included in other translation units, why not replace using-directives with using-declarations like:
using BornAgain::BeamType?

jwuttke: Couldn't we then also say "using std::string, std::vector", and save lots of repetitive "std::" in the bulk of the cpp files? I'd think we are faring quite well with our current explicit style.

waltervh: Yes, we could. Just a matter of style, I suppose. Depending on how extensively a certain type is used in a translation unit, my preference seems to shift.

PR #84, opened 2016-11-24
Alternatives to global functions

jwuttke:
and other mixed edits. This is the final part of the present sequence.


waltervh:
I will postpone review till after group discussion about member vs non-member functions.

jwuttke:
Closed until I restored operators to their out-of-class location.

PR #86, opened 2016-11-25
WavevectorPair aka WavevectorInfo renamed and refactored

jwuttke:
Wavelength removed from member variables, and no longer required as constructor parameter, since it can be easily computed from |K|.


waltervh:
The wavelength cannot be removed from WaveVectorInfo (or Pair) since it refers to the wavelength in vacuum, not in a specific layer. The passed arguments mostly will contain the wavevectors inside a layer.

jwuttke:
Valid point. Request put on hold. Needs more thought.

jwuttke:
Problem with wavelength solved: since the scattering amplitude prefactor only depends on the vacuum wavelength, it can be moved up from FormFactorWrapper to DecoratedLayerComputation.

jwuttke:
Magnetic scattering requires yet another treatment of the trivial factor K^2/4pi.
Improve, or reject this, as you wish, but for the very least, rename wavelength -> vacuumWavelength, or replace it by a precomputed scatteringFactor.

waltervh:
Cherry picked commits are being put on my fork. I will create a merge request from there.

PR #89, opened 2016-11-28
Improved persistence test for Python usage examples (#1609)

jwuttke:
This is an exemplary implementation for one example, namely BuriedParticles. When accepted, it can be very easily extended to all other standard examples (i.e. examples that generate exactly one detector image), and hopefully with little more effort to all other examples.

With this, I propose a slight redesign of the standard example code (see BuriedParticles.py). This makes it easier for the wrapper (see standard_test.py.in) to overwrite the detector settings. For the persistence test, a very small detector grid is sufficient. The choice 5x6 is of course completely arbitrary.


jwuttke:
So I suggest you merge, and we discuss before I extend this solution to other examples.

PR #102, opened 2016-12-01
Coherent sum of form factors now technically possible

waltervh:
The old wrapper now contains a vector of parts that are to be added coherently. Although I did not yet provide any way of adding these parts from the outside, the implementation is now ready to automatically sum up the parts coherently.

### DIFF
--- a/Core/Aggregate/FormFactorCoherentPart.cpp
+++ b/Core/Aggregate/FormFactorCoherentPart.cpp
@@ -21,12 +21,12 @@
 #include "ILayerRTCoefficients.h"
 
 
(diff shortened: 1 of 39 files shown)


waltervh:
By the way, if anyone has a better naming for the coherent sum and its parts, please let me know.

gpospelov:
There is some AppVeyor problem. Appeared in the develop since last merge. It's strange, because pull request itself was build correctly. Let me investigate.

gpospelov:
Can't find solution for Appveyor yet. Temporarily disabled status check. You can merge if you want.

PR #114, opened 2016-12-06
Issues 1667/8: added float sample support to TiffHandler

jburle:
The Desy TIFF files use IEEE float sample format. TiffHandler silently assumed that samples are uint32. Now supports uint32 / int32 / float.

Fixes redmine issues 1667 + 1668.


gpospelov:
Great, thanks!

PR #118, opened 2016-12-08
New INode tree hierarchy

gpospelov:
* New class INode (name can be changed) is introduced between IParameterized and ISample.
* All existing children bookkeeping activity is moved from ISample up to INode.
* ISample::to_str is removed from everywhere in the favor of single INode::to_str()

In the following I'm going to make INode actually own its children, provide unit tests and make some renaming.

### DIFF
--- a/Core/Parametrization/INode.cpp
+++ b/Core/Parametrization/INode.cpp
@@ -7,9 +7,9 @@
 //!
 //! @homepage  http://www.bornagainproject.org
 //! @license   GNU General Public License v3 or higher (see COPYING)
-//! @copyright Forschungszentrum Jülich GmbH 2015
+//! @copyright Forschungszentrum Jülich GmbH 2016
 //! @authors   Scientific Computing Group at MLZ Garching
-//! @authors   C. Durniak, M. Ganeva, G. Pospelov, W. Van Herck, J. Wuttke
+//! @authors   M. Ganeva, G. Pospelov, W. Van Herck, J. Wuttke
 //
 // ************************************************************************** //
 
--- a/Core/Parametrization/INode.cpp
+++ b/Core/Parametrization/INode.cpp
@@ -19,7 +19,7 @@
 #include "NodeUtils.h"
 #include <algorithm>
 
-std::string INode::to_my_str() const
+std::string INode::to_str() const
 {
     return NodeUtils::nodeToString(*this);
 }
--- a/Core/Parametrization/INode.cpp
+++ b/Core/Parametrization/INode.cpp
@@ -16,7 +16,13 @@
 #include "INode.h"
 #include "StringUsageMap.h"
 #include "Exceptions.h"
+#include "NodeUtils.h"
 #include <algorithm>
+
+std::string INode::to_my_str() const
+{
+    return NodeUtils::nodeToString(*this);
+}
 
 void INode::registerChild(INode *sample)
 {
--- a/Core/Parametrization/INode.cpp
+++ b/Core/Parametrization/INode.cpp
@@ -14,3 +14,61 @@
 // ************************************************************************** //
 
 #include "INode.h"
+#include "StringUsageMap.h"
+#include "Exceptions.h"
+#include <algorithm>
+
+void INode::registerChild(INode *sample)
+{
+    if(!sample)
+        throw Exceptions::NullPointerException(
+            "ISample::registerChild -> Error. Null pointer.");
+    m_samples.push_back(sample);
+}
+
+//! remove registered child from the container
+
+void INode::deregisterChild(INode* sample)
+{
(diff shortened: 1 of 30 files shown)


Core/Parametrization/INode.h
```diff
@@ -0,0 +1,57 @@
...
+//
+//! @file      Core/Parametrization/INode.h
+//! @brief     Defines class INode.
+//!
+//! @homepage  http://www.bornagainproject.org
+//! @license   GNU General Public License v3 or higher (see COPYING)
+//! @copyright Forschungszentrum Jülich GmbH 2015
+//! @authors   Scientific Computing Group at MLZ Garching
+//! @authors   C. Durniak, M. Ganeva, G. Pospelov, W. Van Herck, J. Wuttke
+//
```
jwuttke: Let's use 2016 version of header for newly created files.

Core/Parametrization/INode.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "StringUsageMap.h"
+#include "Exceptions.h"
+#include "NodeUtils.h"
+#include <algorithm>
+
+std::string INode::to_str() const
+{
+    return NodeUtils::nodeToString(*this);
+}
+
```
jwuttke: Prefer non-friend function over member function even for the implementation of member function?

Core/Parametrization/INode.h
```diff
@@ -0,0 +1,57 @@
...
+//! @ingroup tools_internal
+
+class BA_CORE_API_ INode : public IParameterized
+{
+public:
+    INode(){}
+    virtual ~INode(){}
+
+    //! Calls the ISampleVisitor's visit method.
+    virtual void accept(ISampleVisitor* p_visitor) const=0;
```
jwuttke: Omit Doxygen comments that are trivial (adding no information to that conveyed by the function name) and repetitive (verbatim repeated in many files), especially in h files (if in doubt, move them to cpp files).

Core/Parametrization/INode.h
```diff
@@ -0,0 +1,57 @@
...
+
+    //! Returns multiline string representing tree structure below the node.
+    virtual std::string to_str() const;
+
+    //! Registers child in the container.
+    void registerChild(INode* sample);
+
```
jwuttke: as above, rm Doxygen comment

Core/Parametrization/NodeUtils.cpp
```diff
@@ -0,0 +1,88 @@
...
+        return result.str();
+    }
+
+    // Returns single line string representing pool parameters of given node.
+    std::string poolToString(const INode &node) {
+        std::ostringstream result;
+
+        const std::vector<RealParameter*> pars = node.getParameterPool()->getParameters();
+        if(pars.size())
+            result << " (";
```
jwuttke: Since the case pars.size()==0 needs special treatment, it would be easier to handle it once and for all at the beginning:
<pre>
    if (pars.size()==0)
        return "";
</pre>

Core/Parametrization/NodeUtils.cpp
```diff
@@ -0,0 +1,88 @@
...
+        const int multiplier = 4;
+        return std::string(multiplier*depth, '.');
+    }
+
+    // Returns multiline string representing pool parameters of given node.
+    std::string poolToMultiString(const INode &node, int depth) {
+        std::ostringstream result;
+
+        for(auto par : node.getParameterPool()->getParameters())
+            result << s_indent(depth)
```
jwuttke: Please use spaces after for and if as per http://apps.jcns.fz-juelich.de/doku/sc/intern:coding_rules

Core/Scattering/ISample.cpp
```diff
@@ -27,79 +27,18 @@ ISample* ISample::cloneInvertB() const
...
-        int ncopy = strUsageMap2[child_name] - 1;
-
-        // If the child is the only one with that name, we do not want any copy number:
-        if (strUsageMap[child_name] == 1)
-            ncopy = -1;
-
-        m_samples[i]->addParametersToExternalPool(new_path, external_pool, ncopy);
+    for(auto child: getChildren() ) {
+        if(const ISample *sample = dynamic_cast<const ISample *>(child)) {
+            for( const IMaterial* material: sample->containedMaterials() )
```
jwuttke: left-align '*'

Core/Parametrization/INode.cpp
gpospelov: In general, why not. For the moment I would like not to populate INode.cpp with lots of code which is related to quite specific use case: printing sample structure from Python. I expect, that INode will get much more functionality related to child/parent relationship, and I think, that printing should be somewhere else. In the future, we might remove  INode::to_str completely.

gpospelov:
I fixed most of requests, and commented on Inode::to_str (above).

PR #119, opened 2016-12-09
plot utilities: repair and improve

jwuttke:
plot_intensity_data_diff.py:
- repair for Python 2.7.12+
- print message and exit if data sets coincide

plot_intensity_data.py:
- optional argument to fix intensity scale.


jwuttke:
one more improvement to come

PR #123, opened 2016-12-14
INode getChildren refactoring

gpospelov:
* INode doesn't own vector of child nodes anymore
* They are generated dynamically on getChildren
* no need to deregisterChild anymore
* Bunch of unit tests to validate getChildren


Core/Aggregate/ParticleLayout.cpp
```diff
@@ -184,6 +184,15 @@ double ParticleLayout::getTotalParticleSurfaceDensity() const
...
 
+std::vector<const INode*> ParticleLayout::getChildren() const
+{
+    std::vector<const INode*> result;
+    for(auto particle : m_particles)
+        result.push_back(particle);
+    result << mP_interference_function;
+    return result;
+}
+
```
jwuttke: A few lines below I see that we have a `registerChild` mechanism. Why then do we need an explicit function `getChildren`? Why not let the registry return all registered children?

Core/Multilayer/LayerInterface.h
```diff
@@ -27,51 +29,51 @@ class BA_CORE_API_ LayerInterface : public ISample
...
-    LayerRoughness *m_roughness;   //!< roughness of the interface
-    const Layer *m_LayerTop;       //!< pointer to the layer above interface
-    const Layer *m_LayerBottom;    //!< pointer to the layer below interface
+    const Layer* m_topLayer;                     //!< pointer to the layer above interface
+    const Layer* m_bottomLayer;                  //!< pointer to the layer below interface
+    std::unique_ptr<LayerRoughness> m_roughness; //!< roughness of the interface
 };
 
+inline const LayerRoughness* LayerInterface::getRoughness() const
+{
```
jwuttke: For me, of the three possible placements of the implementation, this one is the worst.

Core/Aggregate/ParticleLayout.cpp
gpospelov: In our current implementation we have a data duplicated twice. Children are stored on board of INode and on board of ISample. Ownership belongs to ISample. Such approach has following disadvantages:

* Data duplication is always not good.
* The necessity synchronization of two lists of children: i.e. register/deregister mechanism
* From INode it is not clear, that it doesn't own objects, and that copying of INode will give a shallow copy.

Suggested approach is the following. We remove vector of children from INode and make getChildren() pure virtual. This resolves all aforementioned problems. We can now remove register/deregister mechanism completely (althouth, I would like to keep "register" for a while).

PR #131, opened 2017-01-18
FitParameterLinked revision

gpospelov:
Now it is possible to link FitParameterLinked with multiple RealParameters using more than one wildcard based pattern.

* Additionally, FitSuite now reports possible conflicts in fit parameter setup, when one sample parameter is fitted by two different fit parameters.


Core/Fitting/FitParameterLinked.cpp
```diff
@@ -135,3 +137,20 @@ FitParameterLinked::FitParameterLinked(const FitParameterLinked& other)
...
+            return true;
+
+        if(par->getName() != newPar.getName() && par->hasSameData(newPar))
+            throw std::runtime_error("FitParameterLinked::isLinked() -> This is confusing.");
+        if(par->getName() == newPar.getName() && !par->hasSameData(newPar))
+            throw std::runtime_error("FitParameterLinked::isLinked() -> This is confusing.");
+    }
+
+    return  false;
+}
```
waltervh: Can you change this into a clear exception message instead of 'This is confusing'?

gpospelov: In this case
* something really unexpected had happened
* explaining it in details would take too much wording, and probably useless 

I was hesitating between "Panic!", "Bug" and "Should not be here". Please suggest some universal formula.

May be "Unexpected behavior"?

waltervh: What about:
"Different parameter with same data"
"Same parameter name with different data"

waltervh: Shouldn't be long, just clear, in my opinion.

gpospelov: ok

PR #133, opened 2017-01-20
Multipattern fit with FitParameterLinked

gpospelov:
This resolves http://apps.jcns.fz-juelich.de/redmine/issues/1478.

New feature of FitParameterLinked - multi pattern fit - was included in GUI, added to user example and functional test.

FitParameterLinked was renamed to FitParameter.


GUI/coregui/Views/FitWidgets/FitParameterWidget.cpp
```diff
@@ -268,21 +272,22 @@ void FitParameterWidget::initTuningWidgetContextMenu(QMenu &menu)
...
 
-    // --> TODO REDMINE #1478 Uncomment, when issue is solved
-
-//    QStringList fitParNames
-//        = FitParameterHelper::getFitParameterNames(m_jobItem->fitParameterContainerItem());
-//    if(fitParNames.isEmpty() || canCreateFitParameter()==false) {
-//        addToFitParMenu->setEnabled(false);
-//    }
+    const bool allow_one_fit_parameter_to_have_more_than_one_link = true;
+    if (allow_one_fit_parameter_to_have_more_than_one_link) {
```
waltervh: what's the use of this bool definition if it is always the same?

gpospelov: It is temporary, while I'm experimenting with drag-and-drop (equivalent of ifdef/ifndef, comment/uncomment, if you want). Later I will remove this 'if' completely.

waltervh: ok, I will merge then

PR #138, opened 2017-02-02
Making all parameters fittable.

gpospelov:
With this items [Parameter Translation Refactoring](http://apps.jcns.fz-juelich.de/redmine/issues/1626) and [Make All Parameters Fittable](http://apps.jcns.fz-juelich.de/redmine/issues/1624) are basically resolved.

In Core:
* All remaining distributions have been turned to INode.
* More parameters are fittable now (abundance, interference function components, etc).
* Non-relevant parameters are excluded from the fit (like position of the shell, or thickness of the top layer).

In GUI
* Now it is possible to fit in GUI all parameters presented in Core.
* GUITranslation test validates 12 standard sample geometries for correctness of gui->domain translations.
* New translation mechanism provides simpler conversion of GUI names into domain names.

The old translation mechanism is still in place because of ParticleDistributionItem.


Core/Aggregate/ParticleLayout.cpp
```diff
@@ -217,7 +221,9 @@ void ParticleLayout::setAndRegisterInterferenceFunction(IInterferenceFunction* c
 {
     mP_interference_function.reset(child);
     registerChild(child);
-    if(!provideParticleDensity(*mP_interference_function))
+    if(provideParticleDensity(*mP_interference_function))
+        getParameterPool()->removeParameter(BornAgain::TotalParticleDensity);
+    else
         registerParticleDensity();
```
waltervh: Confusing naming here. I first expected that when 'provideParticleDensity' returns true, this density has to be registered. However it is not, because in that case it is provided by the interference function automatically.
Maybe a rename would help here. Something in the lines of 'particleDensityFixedBy(IFF*)'?

gpospelov: Ok, fixed. Waiting for other comments before push.

Tests/Functional/GUI/GUITranslationTest/GUITranslationTest.cpp
```diff
@@ -149,19 +149,11 @@ std::string GUITranslationTest::translationResultsToString() const
...
-        return false;
-
-    size_t layerCount = m_simulation->getSample()->getNumberOfLayers();
-    std::string lastLayerName = "Layer"+std::to_string(layerCount-1)+"/Thickness";
-    if(domainName.find(lastLayerName) != std::string::npos)
-        return false;
-
+    (void)domainName;
     return true;
 }
```
waltervh: Why not remove a method that always returns true?

gpospelov: Before it contained some exceptions to make test passing. Now I fixed all translations, so for the moment it's empty. But it might happen, that on other geometries we will have to put some exceptions again. Probably I will remove it during last cleanup before closing issues 1623, 1624.

waltervh: In principle, code for debugging or temporary hacks should be removed before merges, but since this is in a test, I agree that it doesn't harm so much.

waltervh:
Good work! Nice to see we can use the fitting inside GUI on a lot more parameters as before.

PR #139, opened 2017-02-02
Refactorings for graded layer approach (2)

waltervh:
The Fresnel coefficients have been forwarded deeper into the computation preparation, making it easier now to handle different  R,T coefficients for different particles (or parts thereof). The Computation classes have also been refactored to make the code more concise and easier to maintain or extend.


jwuttke:
Could you please update http://apps.jcns.fz-juelich.de/redmine/projects/bornagain/wiki/GISASSimulation_call_stack so that I can see at a glance how the RT computation has been moved?

waltervh:
The call stack can be easily seen in any debugger, but basically the stack hasn't changed (except for some renaming of methods and a call through the interface of the component computations). Is there a script that produces this call stack?

jwuttke:
Unfortunately not, the wiki page has been generated manually.

gpospelov:
Pull request is OK with me, but I would prefer to see performance comparison, before and after. There was so much unnecessary cloning removed, and I hope for speed up. I can make comparison after the merge, if you want.

waltervh:
Yes Gennady, I would appreciate that. I guess that most runs of the performance tests have been carried out on your system?

gpospelov:
Right, that's why I'm suggesting. So I merge in a minute, if no further comments related to call stack.

gpospelov:
Performance is the same, or marginally better. Merging...

PR #145, opened 2017-02-10
Limits in ParticleDistributionItem

gpospelov:
* RealLimitsItems
* Limits in DistributionItems
* Limits in ParticleDistributionItem, to/from domain propagation
* DistributionWidget was refactored


GUI/coregui/Models/ParticleDistributionItem.cpp
```diff
@@ -96,25 +118,56 @@ void ParticleDistributionItem::updateParameterList()
...
+//    RealLimits limits = linkedItem->limits();
+
+//    auto distr_item = dynamic_cast<DistributionItem*>(
+//                getGroupItem(ParticleDistributionItem::P_DISTRIBUTION));
+//    Q_ASSERT(distr_item);
+
+//    distr_item->setItemValue(DistributionItem::P_IS_INITIALIZED, false);
+//    distr_item->init_parameters(value, limits);
+//}
+
```
waltervh: Should this commented code remain? Same for TestParticleDistributionItem

gpospelov: We have discussed this. Uncommenting the code will lead to the clone() of ParticleDistributionItem being not equal to the original. Still has to be resolved somehow.

PR #146, opened 2017-02-14
WarningSign refactoring

gpospelov:
To unify the usage of WarningSign across the project and reduce code duplication. This resolves [Item 1492](http://apps.jcns.fz-juelich.de/redmine/issues/1492).

### DIFF
--- a/GUI/coregui/Views/InfoWidgets/WarningSign.cpp
+++ b/GUI/coregui/Views/InfoWidgets/WarningSign.cpp
@@ -21,8 +21,8 @@
 #include <QScrollBar>
 
 namespace {
-const int xpos_offset = 38;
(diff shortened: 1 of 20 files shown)


GUI/coregui/Views/InfoWidgets/WarningSign.cpp
```diff
@@ -0,0 +1,131 @@
...
+    , m_warning_header(QStringLiteral("Houston, we have a problem."))
+    , m_warningWidget(0)
+    , m_area(nullptr)
+    , m_clear_just_had_happened(false)
+{
+    setArea(parent);
+}
+
+//! Clears warning message;
+
```
waltervh: Do we want to stick with the "Houston,..." message?

PR #147, opened 2017-02-14
Dynamically determine in which layer a particle sits and use this for ambient material and Fresnel coefficients

waltervh:


Core/Particle/ParticleCoreShell.cpp
```diff
@@ -36,8 +36,6 @@ ParticleCoreShell* ParticleCoreShell::clone() const
 {
     ParticleCoreShell* p_result = new ParticleCoreShell(*mp_shell, *mp_core);
```
gpospelov: It would be nice to change the order of shell/core here too. But will confuse users, I guess.

Core/Scattering/INodeVisitor.h
```diff
@@ -234,7 +233,6 @@ class BA_CORE_API_ INodeVisitor
     virtual void visit(const FormFactorDWBAPol*) {}
     virtual void visit(const FormFactorWeighted*) {}
     virtual void visit(const FormFactorDecoratorDebyeWaller*) {}
-    virtual void visit(const FormFactorDecoratorFactor*) {}
```
gpospelov: FormFactorCoreShell should be declared in visitor too, I guess.

Core/Vector/Transform3D.cpp
```diff
@@ -218,3 +218,29 @@ bool Transform3D::isZRotation() const
     if (m_matrix(2,1) != 0.0) return false;
     return true;
 }
+
+double BottomZ(std::vector<kvector_t> vertices, Transform3D rotation)
```
gpospelov: Const reference to vertices?

Core/Particle/ParticleCoreShell.cpp
waltervh: Yes, I agree with that. Food for discussion? ;-)

Core/Vector/Transform3D.cpp
waltervh: Yes, and Transform3D too.

Core/Scattering/INodeVisitor.h
waltervh: Indeed, I will correct this.

PR #148, opened 2017-02-16
Every form factor can now be queried about its highest/lowest z-coordinate (for graded layer)

waltervh:


gpospelov:
Very nice step forward. We have to think if it's possible to cover all with unit tests later.

PR #149, opened 2017-02-17
JobView refactoring.

gpospelov:
This was about [RefactorJobView and IntensityDataItem](http://apps.jcns.fz-juelich.de/redmine/issues/1490).

* SessionItemWidget was refactored to allow temporary unsubscribtion from item when widget is hidden.
* Most of JobView widgets were switched to it.
* Duplicated ColorMapPlot was removed in the favor of newer ColorMap.
* Old unconvincing projection widget was replaced by placeholder.

To resolve an item completely new ProjectionWidget has to be implemented and it will be done in following request.


waltervh:
Nice refactoring!

PR #152, opened 2017-02-28
Feature #1769

jburle:
The TIFF file from A. Nent has 16b samples. TiffHandler is modified to read 32b floats, and 8, 16, and 32b integer samples.

### DIFF
--- a/Core/InputOutput/TiffHandler.cpp
+++ b/Core/InputOutput/TiffHandler.cpp
@@ -19,16 +19,10 @@
 #include "BornAgainNamespace.h"
 #include "SysUtils.h"
 
-namespace {
-const size_t supported_bitPerSample    = 32;
-const size_t supported_samplesPerPixel = 1;
-const size_t supported_sampleSize      = 4;
-}
-
 TiffHandler::TiffHandler()
     : m_tiff(0)
-    , m_width(0)
-    , m_height(0)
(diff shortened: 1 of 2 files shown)


Core/InputOutput/TiffHandler.cpp
```diff
@@ -107,27 +119,10 @@ void TiffHandler::read_data()
...
-    switch (sampleFormat) {
-    case 1: // unsigned int
-    case 2: // signed int
-    case 3: // IEEE float
-      break;
-    default:
-      throw Exceptions::FormatErrorException("TiffHandler: unexpected sample format");
-    }
-
+    assert(0 == m_bitsPerSample%8);
```
gpospelov: This assert is a bit confusing here. First I thought it is the only check existing, and it is better to replace it with exception throw (to keep GUI alive in the case of unexpected format). Then I saw another check for bitPerSample above, with proper exception throw. So no harm from assert here.

jburle: Yes - that how it was meant! In my code I usually define two macros: EXPECT and ENSURE as assert, and use them to state and document (and debug-time check) preconditions and invariants/postconditions, respectively, instead of 'assert'. Then it is clearer what the purpose is. So EXPECT would be a good choice here, because the condition should be guaranteed upon entry and the next line (/8) relies on it.

PR #161, opened 2017-03-13
Fix bug related to Fresnel coefficients when top layer is not air/vacuum

waltervh:


mganeva:
Changes look good and fix the bug.

PR #166, opened 2017-03-17
GroupProperty refactoring

gpospelov:
New GroupInfo and GroupInfoCatalogue are introduced to provide default type for group item construction (instead of previous alphabetical one).


waltervh:
When adding new files, the updated header should be used, indicating the current year and current developer list.

PR #167, opened 2017-03-20
Usage of average materials for Fresnel calculations can now be enabled in SimulationOptions

waltervh:
Also some refactoring:
-removed 'get' from a number of method names
-replaced inheritance by composition and simplified code of IInterferenceFunctionStrategy and subclasses

And also added a Python example showing the usage of average materials


Core/Computation/MainComputation.h
```diff
@@ -54,11 +55,14 @@ class MainComputation final : public INoncopyable
...
 private:
     void runProtected();
-    static IFresnelMap* createFresnelMap(const MultiLayer* p_multilayer,
-                                         const MultiLayer* p_inverted_multilayer);
+    IFresnelMap* createFresnelMap();
+    // creates a multilayer that contains averaged materials, for use in Fresnel calculations
+    std::unique_ptr<MultiLayer> getAveragedMultilayer();
+    // sets the correct layer materials for the Fresnel map to use
+    void initFresnelMap();
+    bool checkRegions(const std::vector<HomogeneousRegion>& regions) const;
```
gpospelov: May be comment what kind of the check it is, or method renaming.

PR #241, opened 2017-09-15
Redmine feature 1617 (90 deg rotation of 2D OutputData<double>)

DmitryYurov:


DmitryYurov:
revert method will require further adjustments in Axis classes

Core/Binning/ConstKBinAxis.cpp
```diff
@@ -51,6 +51,14 @@ ConstKBinAxis* ConstKBinAxis::clone() const
     return new ConstKBinAxis(getName(), m_nbins, m_start, m_end);
 }
 
+void ConstKBinAxis::revert()
+{
+    VariableBinAxis::revert();
+    double tmp = m_start;
+    m_start = m_end;
+    m_end = tmp;
```
waltervh: Use std::swap instead of manually swapping values.

Core/Instrument/IntensityDataFunctions.cpp
```diff
@@ -60,6 +60,75 @@ OutputData<double>* IntensityDataFunctions::createRelativeDifferenceData(
     return result;
 }
 
+static void rotateDataByOddN(const OutputData<double>& data, OutputData<double>& output,
+    size_t rev_axis_i)
```
waltervh: Prefer unnamed namespaces to static functions for internal linkage (more flexible)

Core/Instrument/IntensityDataFunctions.cpp
```diff
@@ -60,6 +60,75 @@ OutputData<double>* IntensityDataFunctions::createRelativeDifferenceData(
...
+        axis_inds[rev_axis_i] = axis_inds[inv_axis_i];
+        axis_inds[inv_axis_i] = end_bin - tmp_index;
+        size_t output_index = output.toGlobalIndex(std::vector<unsigned>{
+            static_cast<unsigned>(axis_inds[0]), static_cast<unsigned>(axis_inds[1])});
+        output[output_index] = data[index];
+    }
+}
+
+static void rotateDataByEvenN(const OutputData<double>& data, OutputData<double>& output)
+{
```
waltervh: Rename to reflect the fact that this rotates by 180 degrees, not evenN

Core/Instrument/IntensityDataFunctions.cpp
```diff
@@ -60,6 +60,75 @@ OutputData<double>* IntensityDataFunctions::createRelativeDifferenceData(
     return result;
 }
 
+static void rotateDataByOddN(const OutputData<double>& data, OutputData<double>& output,
+    size_t rev_axis_i)
+{
+    const size_t inv_axis_i = 1 ^ rev_axis_i;
+    std::unique_ptr<IAxis> rev_axis(data.getAxis(rev_axis_i).clone());
+    rev_axis->revert();
+    if (rev_axis_i == 1) {
```
waltervh: Instead of cloning and then reverting the axis, use an IAxis method:
IAxis* createReverted() that does both

Core/Instrument/IntensityDataFunctions.cpp
```diff
@@ -60,6 +60,75 @@ OutputData<double>* IntensityDataFunctions::createRelativeDifferenceData(
...
+    std::unique_ptr<OutputData<double>> output(new OutputData<double>());
+    if (n == 1)
+        rotateDataByOddN(data, *output, 1);
+    else if (n == 2)
+        rotateDataByEvenN(data, *output);
+    else
+        rotateDataByOddN(data, *output, 0);
+
+    return output.release();
+}
```
waltervh: Instead of interchanging/reverting axes and then manipulating the data for the three cases separately (consistently with the previous changing of axes), you could:
1) Manipulate the axes according to the required rotation
2) Manipulate the data using the new and old axes directly (without the need for separate cases)

DmitryYurov: Probably, the best option is then using lambda functions to define index mappings. It is worse in the sense of performance, but leave the code readable and enables to move the most part of the code related to transforming 2D array to `rotateDataByN90Deg`

PR #279, opened 2017-11-02
SessionItemWidget refactoring

gpospelov:
SessionItemWidget is a common base intended for SessionItem presentation. It's main feature was subscription/unsubscription mechanism to allow update on SessionItem's change, when widget is visible.
This mechanism was also responsible for the time of life: widget is destroyed before item, item is destroyed before widget.

* With this pull request, the logic is moved out of widget to the dedicated SessionItemController.
* Lots of unit tests for the controller
* Subtle bug with access to the deleted widget was caught

### DIFF
--- a/GUI/coregui/Views/CommonWidgets/SessionItemController.cpp
+++ b/GUI/coregui/Views/CommonWidgets/SessionItemController.cpp
@@ -81,7 +81,9 @@ void SessionItemController::unsubscribe()
     if (!m_item)
         return;
 
-    unsubscribeParent();
+    if (m_parent_subscribed)
+        unsubscribeParent();
+
     m_item->mapper()->unsubscribe(parent());
 }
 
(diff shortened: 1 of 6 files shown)


GUI/coregui/Views/CommonWidgets/SessionItemController.cpp
```diff
@@ -0,0 +1,118 @@
...
+    onControllerDestroy();
+}
+
+void SessionItemController::setItem(SessionItem* item)
+{
+    if(m_item == item)
+        return;
+
+    if (m_item) {
+        m_item->mapper()->unsubscribe(this);
```
DmitryYurov: Unsubscription is done two times here, firstly through SessionItem, secondly through SessionItemController method. May be, it is worth changing in future commits?

gpospelov: There are two actors here. One is a controller which is subscribed to the item to know about it's destruction or to warn item about destruction of itself. Another actor is parent, who can be subscribed to lots of other item's signals. When widget (parent) gets hidden, it gets unsubscribed from item by the controller. But controller itself stays subscribed. Method SessionItemController::subscribe is for the parent only.

PR #281, opened 2017-11-03
Improved material class hierarchy and added compile-time typization of material classes

DmitryYurov:


DmitryYurov:
I will close commit for now. Need some time to figure out what happens on windows platform

PR #284, opened 2017-11-06
Projections can be exported to ASCII file

gpospelov:


DmitryYurov:
May be it is a good idea to make a functional test for the export to ascii in the future?

PR #285, opened 2017-11-06
Redmine issue 1876: vacuum material implementation

DmitryYurov:


Core/Material/VacuumMaterial.cpp
```diff
@@ -0,0 +1,32 @@
+#include "VacuumMaterial.h"
```
gpospelov: I think we do not need dedicated class for VacuumMaterial. Vacuum can be already created with two other methods.

DmitryYurov: It's true, it can. But then it is not clear how to make averaging of materials in layer, given WavelengthIndependentMaterial and RefractiveCoefMaterial must not be used simultaneously

DmitryYurov:
Will need to add some more tests and fix a bug

Core/Material/VacuumMaterial.cpp
gpospelov: If there is a mixture of material types, we throw an exception. I found confusing to have yet another way to do the same. Vacuum is not that different from other materials to have dedicated class.

PR #290, opened 2017-11-07
Redmine issue 1871: implement universal material averaging

DmitryYurov:


DmitryYurov:
Last commit handles another redmine issue, but since the issue is very small, there is no sense to create separate pull request

PR #295, opened 2017-11-13
Small refactoring of SessionItem

waltervh:


GUI/coregui/Models/SessionItemUtils.cpp
```diff
@@ -25,6 +25,11 @@ int SessionItemUtils::ParentRow(const SessionItem& item)
     return -1;
 }
 
+bool SessionItemUtils::HasCHildren(const SessionItem& item)
```
gpospelov: I would not move hasChildren to utils for the sake of similarity of our SessionItem and QStandardItem. QStandardItem has a method hasChildren

GUI/coregui/Views/InstrumentWidgets/BeamEditorWidget.cpp
```diff
@@ -59,12 +59,12 @@ BeamEditorWidget::BeamEditorWidget(QWidget* parent)
     mainLayout->addStretch();
     setLayout(mainLayout);
 
-    connect(m_wavelengthPresenter, SIGNAL(onDialogRequest(SessionItem*, QString)),
```
gpospelov: While fully supporting an idea of using new signal/slot syntax, I would switch to it only while making some other (bigger) refactoring in a widget.

GUI/coregui/Models/SessionItemUtils.cpp
waltervh: Right, I will move it back

GUI/coregui/Views/InstrumentWidgets/BeamEditorWidget.cpp
waltervh: I did it here to remove the Q_DECLARE_METATYPE with its obscure comment

gpospelov: ah, ok

PR #298, opened 2017-11-14
Redmine issue 1884: minor improvements in SpecularSimulation

DmitryYurov:


Core/Simulation/SpecularSimulation.cpp
```diff
@@ -18,32 +18,32 @@
...
 #include "SpecularMatrix.h"
 #include "MaterialUtils.h"
+#include "Histogram1D.h"
 #include <memory>
 
 SpecularSimulation::SpecularSimulation() : Simulation()
 {
-    initialize();
+    setName(BornAgain::SpecularSimulationType);
 }
```
waltervh: The 'initialize()' method localized the necessary steps (only one here) for construction of an object. Please do not replace explicit initialization.

Core/Simulation/SpecularSimulation.h
```diff
@@ -39,45 +40,65 @@ class BA_CORE_API_ SpecularSimulation : public Simulation
...
 
-    //! returns vector of transmission coefficients for all alpha_i angles for given layer index
+    //! Returns vector of transmission coefficients for all alpha_i angles for given layer index.
+    //! Method is deprecated in favor of SpecularSimulation::transmissivity.
     std::vector<complex_t> getScalarT(size_t i_layer) const;
 
-    //! returns vector of Kz coefficients for all alpha_i angles for given layer index
+    //! Returns vector of Kz coefficients for all alpha_i angles for given layer index.
+    //! Method is deprecated and will be removed in next release.
     std::vector<complex_t> getScalarKz(size_t i_layer) const;
```
waltervh: Only the reflectivity of the top layer is needed. We do not yet have any requests or applications where transmission is needed or reflectivity in other layers is needed.
getScalarR/T/Kz are not deprecated, since these are necessary for calculation of depth intensity profiles (used by Marina for a project).

gpospelov: Do we have a script where she plots depth profile? It will give us idea how to change API. It would be nice to embed depth profile access to SpecularSimulation class.

Core/Simulation/SpecularSimulation.h
```diff
@@ -39,45 +40,65 @@ class BA_CORE_API_ SpecularSimulation : public Simulation
...
+    //! Consider using IHistogram::getXaxis.
     const IAxis* getAlphaAxis() const;
 
-    //! returns vector of reflection coefficients for all alpha_i angles for given layer index
+    //! Returns reflectivity values \f$Reflectivity = \|R\|^2\f$ from the upper layer in the form of
+    //! OutputData<double>.
+    virtual OutputData<double>* getDetectorIntensity(IDetector2D::EAxesUnits units_type
+                                                     = IDetector2D::DEFAULT) const override;
+
+    Histogram1D* reflectivity() const {return reflectivity(0);}
```
gpospelov: I was stacked on this line for a minute trying to understand why thw default parameter is not used.

PR #306, opened 2017-11-21
Finished extracting IDetector and employed it in simulations

DmitryYurov:


Core/Simulation/GISASSimulation.cpp
```diff
@@ -84,35 +84,49 @@ void GISASSimulation::setDetector(const IDetector2D& detector)
...
+IDetector2D* invoke2DDetector(Instrument& instrument)
+{
+    IDetector2D* detector = instrument.detector2D();
+    if (!detector)
+        throw std::runtime_error(
+            "Error in GISASSimulation: wrong detector type");
+    return detector;
+}
+}
+
```
waltervh: Please rename this function at some point. It doesn't invoke anything, but retrieves a 2d detector if there is one. Suggestion: Get2DDetector(...)

waltervh: Or probably better: Detector2D(...)

PR #310, opened 2017-11-23
Redmine issue 1902: basic implementation of a detector for specular simulations

DmitryYurov:


Core/Instrument/SpecularDetector1D.cpp
```diff
@@ -24,3 +29,83 @@ class SpecularPixel : public IPixel
...
+        result.push_back(std::move(sim_element));
+    }
+
+    return result;
+}
+
+std::string SpecularDetector1D::axisName(size_t index) const
+{
+    if (index == 0) {
+        return BornAgain::U_AXIS_NAME;
```
waltervh: Why is it U_AXIS_NAME?

PR #312, opened 2017-11-27
Add constant background object to Simulation for fitting

waltervh:


Core/Simulation/Simulation.h
```diff
@@ -102,6 +103,11 @@ class BA_CORE_API_ Simulation : public ICloneable, public INode
     //! Update the sample by calling the sample builder, if present
     void updateSample();
 
+    //! Generate a single threaded computation for a given range of SimulationElement's
+    virtual std::unique_ptr<MainComputation> generateSingleThreadedComputation(
```
DmitryYurov: But why should it be virtual?

DmitryYurov: Ok, understood

PR #313, opened 2017-11-27
ScientificDoubleProperty is removed

gpospelov:
Dedicated QVariant property to handle value of 'double' in scientific way is removed.
* Now every double can be shown in scientific notation without breaking back compatibility
* To demonstrate this, right-mouse-button context menu is implemented for SampleView's property editor, where one can change scientific notation and number of decimals.

### DIFF
--- a/GUI/coregui/Models/SessionModelDelegate.cpp
+++ b/GUI/coregui/Models/SessionModelDelegate.cpp
@@ -28,11 +28,6 @@ bool isDoubleProperty(const QModelIndex& index)
     return index.data().type() == QVariant::Double;
 }
 
-bool isIntProperty(const QModelIndex& index)
-{
-    return index.data().type() == QVariant::Int;
-}
-
 //! Returns text representation of double value depending on user defined editor type.
 //! FIXME Remove this temporary function after getting rid from ScientificDoubleProperty
 QString doubleToString(const SessionItem& item)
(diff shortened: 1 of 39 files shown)


GUI/coregui/Models/item_constants.h
```diff
@@ -269,6 +269,9 @@ const ModelType ComboPropertyType = "ComboProperty";
 const ModelType ColorPropertyType = "ColorProperty";
 const ModelType GroupPropertyType = "GroupProperty_t";
 
+// --- Custom editors for variant propertues ---
+const ModelType DefaultEditorType = "Default";
+const ModelType ScientificEditorType = "SceintificDouble";
```
DmitryYurov: Typo here (just in case it makes difference)

GUI/coregui/Models/SessionModelDelegate.cpp
```diff
@@ -28,9 +28,21 @@ bool isDoubleProperty(const QModelIndex& index)
     return index.data().type() == QVariant::Double;
 }
 
-bool isIntProperty(const QModelIndex& index)
+//! Returns text representation of double value depending on user defined editor type.
+//! FIXME Remove this temporary function after getting rid from ScientificDoubleProperty
```
DmitryYurov: Do I understand properly, that this function will be removed in further commits? Seems that ScientificDoubleProperty has been removed.

PR #321, opened 2017-11-30
Provide combobox for background selection

waltervh:
I also added PoissonNoiseBackground as an exercise to see how easy it is to add new background objects to the core and GUI (and Python)


Core/Tools/MathFunctions.cpp
```diff
@@ -370,15 +370,32 @@ double MathFunctions::GenerateUniformRandom()
...
 double MathFunctions::GenerateNormalRandom(double average, double std_dev)
 {
     return GenerateStandardNormalRandom()*std_dev + average;
 }
 
-double MathFunctions::GenerateStandardNormalRandom() // using c++11 standard library
+double MathFunctions::GeneratePoissonRandom(double average) // using c++11 standard library
 {
     unsigned seed = static_cast<unsigned>(std::chrono::system_clock::now().time_since_epoch().count());
     std::default_random_engine generator(seed);
```
gpospelov: In GUIPerformanceTest I'm using std::random_device and mt19937. We have to agree at some point on common way...

PR #324, opened 2017-12-01
Redmine issue #1910: adapting standard computational means to use with SpecularSimulation

DmitryYurov:


DmitryYurov:
For some reason relative difference in results for the new RoughnessInSpecular functional test is above standard threshold (1e-9 vs 1e-10). Should be investigated further, muted for now.

DmitryYurov:
This discrepancy can be observed with only enabling the new test, with or without other changes of the pull-request

PR #327, opened 2017-12-04
Redmine issue #1910: Finally resetting SpecularSimulation to the usage of Specular methods and data

DmitryYurov:


DmitryYurov:
typo: using of _Simulation_ methods and data

PR #330, opened 2017-12-06
Lattice.h Unit Test

anikhalder:
@waltervh Can you please check the reciprocalLatticeVectorsWithinRadius() method in Lattice.h ?


Tests/UnitTests/Core/utils/google_test.h
```diff
@@ -0,0 +1,16 @@
...
+#pragma warning ( push )
+#pragma warning ( disable: 4275 )
+#include <gtest/gtest.h>
+#pragma warning ( pop )
+#else
+#include <gtest/gtest.h>
+#endif
+
+
+#endif //GOOGLE_TEST_H
```
waltervh: Good idea to extract this common piece of code from the unit test, but now there is a mixed usage. I suggest to put this header in UnitTests/utilities folder, remove the extra include directory from CMakeLists.txt and make the other unit test files conform to the same style (using the common include).

PR #338, opened 2017-12-11
Minor changes to BornAgain

DmitryYurov:
1. Fixed the problem with printing error messages from C++-derived classes (e.g. DefaultFitObserver)
2. Introduced another convenience factor "barn" (1e-10 nm^2)
2. Added another test for specular simulation. Currently a discrepancy between genx and BornAgain results is observed for this test


DmitryYurov:
At last builds on all platforms

PR #353, opened 2018-01-04
Redmine issue 1917: std::vector<SimulationElement> replaced with templates

DmitryYurov:
In the future it will help to use several different types of simulation elements without performance decrease

### DIFF
--- a/Core/Computation/DWBAComputation.cpp
+++ b/Core/Computation/DWBAComputation.cpp
@@ -15,7 +15,6 @@
 #include "DWBAComputation.h"
 #include "ParticleLayoutComputation.h"
 #include "Layer.h"
-#include "IFresnelMap.h"
 #include "MatrixFresnelMap.h"
 #include "MultiLayer.h"
 #include "RoughMultiLayerComputation.h"
@@ -30,11 +29,7 @@ static_assert(std::is_copy_constructible<DWBAComputation>::value == false,
 static_assert(std::is_copy_assignable<DWBAComputation>::value == false,
     "DWBAComputation should not be copy assignable");
 
-DWBAComputation::DWBAComputation(const MultiLayer& multilayer, const SimulationOptions& options,
-                                 const std::shared_ptr<ProgressHandler>& progress,
-                                 const std::vector<SimulationElement>::iterator& begin_it,
-                                 const std::vector<SimulationElement>::iterator& end_it)
-    : IComputation(options, progress, begin_it, end_it, multilayer)
+void DWBAComputation::init()
 {
     mP_fresnel_map = createFresnelMap();
     bool polarized = mP_multi_layer->containsMagneticMaterial();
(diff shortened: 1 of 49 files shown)


Core/Simulation/Simulation.cpp
```diff
@@ -142,10 +142,9 @@ void Simulation::runSimulation()
     for (size_t index = 0; index < param_combinations; ++index) {
         double weight = m_distribution_handler.setParameterValues(P_param_pool.get(), index);
         runSingleSimulation();
-        addElementsWithWeight(m_sim_elements.begin(), m_sim_elements.end(), total_intensity.begin(),
-                              weight);
+        addElementsWithWeight(total_intensity.begin(), weight);
```
gpospelov: I liked previous signature more because it reminds standard signature in stl's algorithms

Core/Computation/DWBAComputation.cpp
```diff
@@ -31,7 +31,7 @@ static_assert(std::is_copy_assignable<DWBAComputation>::value == false,
     "DWBAComputation should not be copy assignable");
 
 DWBAComputation::DWBAComputation(const MultiLayer& multilayer, const SimulationOptions& options,
-                                 ProgressHandler& progress,
+                                 const std::shared_ptr<ProgressHandler>& progress,
```
gpospelov: So far we were using shared_ptr only where it was really unavoidable. I would not start to use it without real necessity

Core/Simulation/Simulation.cpp
DmitryYurov: Since addElementsWithWeight has become a private method of SimulationImpl (which is a template), it will require additional template parameters to save previous function signature.

Core/Computation/IComputation.h
```diff
@@ -47,11 +48,29 @@ class IComputation
     std::string errorMessage() const { return m_status.errorMessage(); }
 
 protected:
+    struct IIterHandler
```
gpospelov: It's hard to judge at this point whether IIterHandler is an elegant replacement of just  two begin,end pointers

Core/Simulation/GISASSimulation.h
```diff
@@ -25,7 +26,7 @@ class Histogram2D;
 //! Main class to run a Grazing-Incidence Small-Angle Scattering simulation.
 //! @ingroup simulation
 
-class BA_CORE_API_ GISASSimulation : public SimulationImpl
+class BA_CORE_API_ GISASSimulation : public SimulationImpl<SimulationElement>
```
gpospelov: I vote for composition (templated member responsible for production of vector of certain SimulationElements) instead of templated inheritance.

gpospelov:
My general filling is that we have to invest a bit more in the design. I think, Simulation base class should contain only very basic methods (e.g. ::runSimulation()) and shouldn't be templated. Derived classes should contain composition, something like GISASKernel<ParticularSimElement>, responsible for creation of vector of SimulationElement and particular computation terms. The knowledge about SimulationElements should be removed from IDetector.

PR #355, opened 2018-01-12
Redmine issue 1917: removed simulation elements from Simulation class

DmitryYurov:
1. Iterator-based methods replaced by index-based methods in Simulation
2. Simulation elements moved down to leaf classes in Simulation hierarchy

Seems that part of Simulation functionality can be isolated in a separate hierarchy with a templated class

### DIFF
--- a/Core/Simulation/SimElementUtils.h
+++ b/Core/Simulation/SimElementUtils.h
@@ -1,0 +1,22 @@
+// ************************************************************************** //
+//
+//  BornAgain: simulate and fit scattering at grazing incidence
+//
+//! @file      Core/Simulation/SimElementUtils.h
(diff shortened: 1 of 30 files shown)


Core/Simulation/SimElementUtils.h
```diff
@@ -0,0 +1,22 @@
...
+// ************************************************************************** //
+
+#ifndef SIMELEMENTUTILS_H_
+#define SIMELEMENTUTILS_H_
+
+namespace SimElementUtils {
+
+}
+
+#endif /* SIMELEMENTUTILS_H_ */
```
waltervh: Just as a remark: we generally do not abbreviate names in the code

PR #361, opened 2018-01-16
Redmine issue 1931: Footprint correction for specular simulation

DmitryYurov:
Not functional for now

### DIFF
--- a/Core/Instrument/GaussianBeamFormFactor.cpp
+++ b/Core/Instrument/GaussianBeamFormFactor.cpp
@@ -1,0 +1,13 @@
+#include "GaussianBeamFormFactor.h"
+#include "MultiLayer.h"
+
+GaussianBeamFormFactor::GaussianBeamFormFactor(double width) : m_beam_width(width)
+{
+}
+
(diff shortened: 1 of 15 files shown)


gpospelov:
Why not to localize all in single class - FootprintCorrection. There can be characteristic sample size, beam shape and necessary calculations. In my opinion there is no strong need to couple it with MultiLayer.

DmitryYurov:
I don't have objections except that FootprintCorrection is an abstract entity. Assigning beam shape (a better name than BeamFormFactor, by the way) to the beam and a dimension to the sample is more natural from user point of view.

DmitryYurov:
You are right of course, that adding one more parameter to MultiLayer does no good. FootprintCorrection class would enable us to avoid extension of MultiLayer class.

gpospelov:
I don't mind to proceed like it is now. We can change it later, if necessary. I don't know what will be more natural from user point of view. But from GUI perspective, having all numbers responsible for footprint in one place (InstrumentWidget), make it easier to tune.

DmitryYurov:
Ok, I will have it in mind in course of future work.

PR #364, opened 2018-01-16
Fix Boost include for GUI

tschoonj:
The following fix is necessary to ensure that the Boost installation as
found by CMake will be used, instead of the one that is installed in
/usr/include.


waltervh:
Thank you for the fix! However, our master branch is frozen, except for critical bugfixes. So if you could create the pull request on the develop branch, we would be happy to accept it.

tschoonj:
Ok, opening a new PR

PR #365, opened 2018-01-17
Fix Boost include for GUI

tschoonj:
The following fix is necessary to ensure that the Boost installation as
found by CMake will be used, instead of the one that is installed in
/usr/include.


waltervh:
Seems like they are having some issues on Travis CI. We will merge as soon as they have this fixed.

tschoonj:
Thanks!

PR #371, opened 2018-01-24
Redmine issue 1860: getting ready for beam resolution effects in SpecularSimulation

DmitryYurov:
The following changes were made:
1. Replaced std::vector\<SpecularSimulationElement\> cache with std::vector\<double\> cache
2. Moved getScalar(R/T/Kz) to private section (with the new cache these methods would not work) and got rid of their usage in examples.
3. yaml_utils.py is not anymore used in PyPersistence to write down data. SpecularSimulation results are now written into *.int.gz.
4. Fixed a bug related to copying genX data into installation directory


gpospelov:
Appveyor failure was caused by newer Qt they started to use, see [here](https://www.appveyor.com/docs/build-environment/). Fix in .appveyor.yml is required.

gpospelov:
Proposed changes will make our 'make fullcheck' to failure. Please investigate functional tests in Functional/Python/PyPersistence

PR #376, opened 2018-01-30
Redmine issue 1860: beam effects now in operation in specular simulations

DmitryYurov:


Core/Parametrization/INode.cpp
```diff
@@ -88,15 +88,15 @@ std::string INode::displayName() const
...
 
     NodeIterator<PreorderStrategy> it(this);
     it.first();
     while (!it.isDone()) {
         const INode *child = it.getCurrent();
-        const std::string path = NodeUtils::nodePath(*child, this->parent()) + "/";
+        const std::string path = NodeUtils::nodePath(*child, node) + "/";
         child->parameterPool()->copyToExternalPool(path, result.get());
         it.next();
     }
```
waltervh: There is no need to change the interface of INode::createParameterTree, since it does exactly what it should do.

PR #433, opened 2018-03-06
Redmine issue 1974: beam resolution functionality for specular simulations in GUI

DmitryYurov:


waltervh:
In EnvironmentEditor.h, all Simulation types are now accepted, but the comment still refers only to GISASSimulation/

PR #457, opened 2018-03-28
Preliminary changes to Fresnel maps interface and fix for memory leaks

DmitryYurov:
1. IFresnelMap::getInCoefficients now accepts any type of simulation elements (will be required for the future depth probe machinery)
2. getInCoefficients/getOutCoefficients now operate on unique pointers. Besides, FormFactors now own LayerCoefficients. On the one hand, it can lead to overhead due to copying LayerCoefficients from one FormFactor object to another , but, on the other hand, code is very prone to memory leaks and segfaults otherwise. At least one memory leak has been already found and fixed. If copying overhead is critical, I would propose using a shared pointer.

Leakage was at 40th line in GISASSpecularComputationTerm.cpp


gpospelov:
After some thought I can't say that I'm fully happy with this already accepted pull request.

From one side, making ScalarFresnelMap returning unique_ptr was probably a good idea. From another side, making FormFactor own unique_ptr requires lots of copying all the way day in formfactor hierarchy. 

Previous design with naked const_ptr was to my opinion justified. Memory management was done by the parent object,  while formfactor itself was providing low level calculations.

PR #467, opened 2018-04-11
Update RealSpaceDialog persistent settings and implement Lock view checkbox

anikhalder:
The following have been attempted:

- Implement persistent settings of RealSpaceDialog window.
- Implement 'Lock view' checkbox in RealSpaceToolbar that allow users to keep the current view of their selection.
- Synchronize the Canvas view with the selection/unselection and deletion of items.


GUI/coregui/Views/RealSpaceWidgets/RealSpaceCanvas.cpp
```diff
@@ -25,13 +25,17 @@ RealSpaceCanvas::RealSpaceCanvas(QWidget* parent)
...
     , m_view(new RealSpaceView)
+    , m_lockView_on(false)
 {
     QVBoxLayout* layout = new QVBoxLayout;
     layout->setMargin(0);
     layout->setSpacing(0);
     layout->addWidget(m_view);
     setLayout(layout);
 
+    // listening whether lock view in RealSpaceToolBar has been unchecked or not
```
gpospelov: Do you need a signal here?

GUI/coregui/Views/RealSpaceWidgets/RealSpaceCanvas.cpp
```diff
@@ -99,6 +108,18 @@ void RealSpaceCanvas::onFaceViewAction()
     faceView();
 }
 
+void RealSpaceCanvas::onLockViewAction(bool lockView_on)
+{
+    // if lock view is unchecked, emit a signal to display the current selection on the canvas
+    if(m_lockView_on == true && lockView_on == false)
+    {
+        m_lockView_on = lockView_on;
+        emit lockViewUnchecked(m_selectionModel->selection(), m_selectionModel->selection());
```
gpospelov: It's a bit misleading to send selection() twice to the function which expected "unselected" as a second parameter.
Better make new method updateToSelection() and use it everywhere.

gpospelov:
Thanks!

PR #470, opened 2018-04-13
Provide uniform access interface for simulation/fitting results (redmine #2014)

gpospelov:
* SimulationArea is hidden inside IDetector, others using IDetector::iterate
* FitSuite switched to SimulationResult
* FitObject now is more robust against real data shape (will make it easier to set data from GUI)


Core/Instrument/IDetector.h
```diff
@@ -37,6 +38,7 @@ class RegionOfInterest;
 class BA_CORE_API_ IDetector :  public ICloneable, public INode
 {
 public:
+    using const_iterator = const SimulationAreaIterator&;
```
DmitryYurov: SWIG is known to experience problems with using directives. Should be tested and replaced by typedef if necessary

gpospelov: Up to now we didn't have any strong policy related to what we are exposing to SWIG and how we test what have been exposed. "Minimum Python API" is only in far feature plans.

* So far, we didn't provide special testing for Python bindings.
* Binding was always considered as successful, unless there is no complains from SWIG and code compiles on our build server.

Shall I start to test bindings already from this pull request and/or start to populate code with ifndef SWIG already now?

Or, maybe, may I suggest to postpone this discussion until a special occasion and, meanwhile, have this pull request accepted? Would it be an option?

DmitryYurov: Amendment: aliasing is somewhat tricky in the case of templates. Anyway, it is a minor issue. I am merging anyway

PR #480, opened 2018-04-18
Enabled mouse tracking in specular plots

DmitryYurov:


gpospelov:
Another possibilities

* Do not introduce any strategies into PlotEventInfo, just add bool/enum which will change the behavior of status string, depending of 1d or 2d context. This will save couple of hundreds of LOC.

* Introduce new class ScientificPlot, which will contains via composition QCustomPlot and will be used, via composition again, in SpecularPlot and ColorMap. Functionality of ColorMapEvent can be then merged into ScientificPlot.

PR #491, opened 2018-04-27
Refactor RealSpaceBuilder, implement SceneGeometry class and enlarge/reduce layer size RealSpaceToolbar buttons

anikhalder:


GUI/coregui/Views/RealSpaceWidgets/RealSpaceCanvas.cpp
```diff
@@ -26,6 +26,7 @@ RealSpaceCanvas::RealSpaceCanvas(QWidget* parent)
     , m_sampleModel(nullptr)
     , m_view(new RealSpaceView)
     , m_view_locked(false)
+    , m_sceneGeometry(new SceneGeometry)
```
gpospelov: I think it is not deleted. At the next occasion, please wrap it in unique_ptr.

anikhalder: Thank you very much for pointing that out. I have wrapped it up into a unique pointer. The next pull request will contain the change.

PR #492, opened 2018-05-02
Implement rotation of particles and locking of camera orientation on LockView

anikhalder:


GUI/coregui/Views/RealSpaceWidgets/RealSpaceBuilderUtils.cpp
```diff
@@ -298,3 +301,42 @@ QVector<QVector<double>> RealSpaceBuilderUtils::computeInterference2DLatticePosi
...
+    float yaw = 0.0f;
+    float roll = 0.0f;
+
+    auto transformationItem = particleItem.getItem(ParticleItem::T_TRANSFORMATION);
+
+    auto rotItem = transformationItem->getGroupItem(TransformationItem::P_ROT);
+
+    if(rotItem->modelType() == Constants::ZRotationType)
+    {
+        auto ZRotItem = static_cast<ZRotationItem*>(rotItem);
```
waltervh: I think there's no need for the casts here. If it is a ZRotationItem, it will have the P_ANGLE child item. Same for the other ones.

anikhalder: Thanks. I made the changes.

PR #512, opened 2018-05-23
GUI ParticleComposition implementation

anikhalder:
Temporarily in  files MaterialPropertyController and TestMaterialPropertyController the connection for propagating signal (dataChanged and rowRemoval) between original material model and the material property's model has been disabled.

### DIFF
--- a/Core/DecoratedFormFactor/IFormFactorDecorator.h
+++ b/Core/DecoratedFormFactor/IFormFactorDecorator.h
@@ -48,7 +48,7 @@ public:
     double topZ(const IRotation& rotation) const override {
         return mp_form_factor->topZ(rotation); }
 
-    IFormFactor* getFormFactor() const {
+    const IFormFactor* getFormFactor() const {
         return mp_form_factor; }
 
 protected:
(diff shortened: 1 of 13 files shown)


Core/DecoratedFormFactor/IFormFactorDecorator.h
```diff
@@ -48,6 +48,9 @@ class BA_CORE_API_ IFormFactorDecorator : public IFormFactor
     double topZ(const IRotation& rotation) const override {
         return mp_form_factor->topZ(rotation); }
 
+    IFormFactor* getFormFactor() const {
+        return mp_form_factor; }
+
```
waltervh: can you make this method return a const IFormFactor*?

GUI/coregui/Views/RealSpaceWidgets/TransformTo3D.cpp
```diff
@@ -23,6 +23,30 @@
...
+#include "FormFactorHemiEllipsoid.h"
+#include "FormFactorIcosahedron.h"
+#include "FormFactorPrism3.h"
+#include "FormFactorPrism6.h"
+#include "FormFactorPyramid.h"
+#include "FormFactorTetrahedron.h"
+#include "FormFactorTruncatedCube.h"
+#include "FormFactorTruncatedSphere.h"
+#include "FormFactorTruncatedSpheroid.h"
+
```
waltervh: This can be replaced with:
#include HardParticles.h

waltervh:
Nice work! If you can take into account my two comments, I will merge.

anikhalder:
Thanks a lot! I implemented the changes.

PR #526, opened 2018-05-30
GUI3D ParticleCoreShell implementation

anikhalder:


GUI/coregui/Views/RealSpaceWidgets/RealSpaceBuilderUtils.cpp
```diff
@@ -351,6 +353,47 @@ void RealSpaceBuilderUtils::applyParticleTransformations(const Particle* particl
...
+        float y = static_cast<float>(particle->position().y());
+        float z = static_cast<float>(particle->position().z());
+
+        // position of the whole particleCoreShell
+        kvector_t positionCoreShell = particleCoreShell->position();
+
+        RealSpace::Vector3D position(x + origin.x() + static_cast<float>(positionCoreShell.x()),
+                                     y + origin.y() + static_cast<float>(positionCoreShell.y()),
+                                     z + origin.z() + static_cast<float>(positionCoreShell.z()));
+
```
waltervh: This is not correct. You have to explicitly apply the rotation and translation to the particle and then retrieve rotation and position.

anikhalder: Implemented the correction. Hope it should work now.

PR #536, opened 2018-06-15
GUI Refactor interference functions (2D and 1D lattice) by using core objects

anikhalder:


GUI/coregui/Views/RealSpaceWidgets/RealSpaceBuilder.cpp
```diff
@@ -149,20 +149,20 @@ void RealSpaceBuilder::populateInterference(RealSpaceModel* model,
     else if (interferenceLattice->modelType() == Constants::InterferenceFunction1DLatticeType)
     {
         auto interferenceItem =
-                dynamic_cast<InterferenceFunction1DLatticeItem *>(interferenceLattice);
+                dynamic_cast<const InterferenceFunction1DLatticeItem *>(interferenceLattice);
         auto interference = interferenceItem->createInterferenceFunction();
         RealSpaceBuilderUtils::populateInterference1DLatticeType(interference.get(), model,
                                                                  layoutItem, sceneGeometry, this);
```
waltervh: You could make this a bit simpler by dynamic_casting to "const InterferenceFunctionItem*", because this is a common base class that supports createInterferenceFunction()

anikhalder: Thanks! I've changed it accordingly.

PR #549, opened 2018-06-29
FitObjective machinery as a replacement of FitSuite

gpospelov:
This resolves [Introduce SimulationBuilder and refactor FitSuite accordingly](http://apps.jcns.fz-juelich.de/redmine/issues/2079).

* FitObjective and SimDataPair fully replace the whole FitSuite* family
* From now, simulation in fitting should be constructed using python callable and set of parameters from minimizer.
* See FitCylinderPrisms_new.py which demonstrates the idea.


waltervh:
Closing because of failing build on buildbot CentOS configuration

PR #555, opened 2018-07-06
Redmine issue 2106: using DataItem1DView in fit view for 1D data

DmitryYurov:
1. DataItem1DView does not change the state of DataItem anymore.
2. Connection of DataItem1DView to the instrument (for units conversion) is now done through finding parenting JobItem and using its InstrumentItem.
3. Unit conversion is done "on the fly", when graph data is requested, i.e. DataItem1DView does not store output data


DmitryYurov:
Thanks!

PR #559, opened 2018-07-12
Generate debian package using cpack

juanmcloaiza:
To generate a debian package with cpack pass "-D BUILD_DEBIAN=ON" to cmake


cmake/bornagain/modules/CPackDebian.cmake
```diff
@@ -19,12 +19,12 @@ IF(NOT CPACK_DEBIAN_PACKAGE_ARCHITECTURE)
...
+set(CPACK_DEBIAN_PACKAGE_DEPENDS "build-essential, git, cmake(>=3.1), libgsl-dev(>=1.15), libboost-all-dev(>=1.48), libfftw3-dev(>=3.3.1), python3, python3-dev, python3-numpy, python3-matplotlib, libtiff5-dev(>=4.0.2), qt5-default(>=5.4), libqt5designercomponents5, qttools5-dev, libqt5svg5-dev")
```
mganeva: you don't need build-essential, git, cmake, etc to run the debian package. Please, clean up the list of dependencies.

mganeva:
Try `or` statements for different versions of the boost libraries (see https://stackoverflow.com/questions/2139124/how-to-make-a-debian-package-depend-on-multiple-versions-of-libboost)

mganeva:
The debian package should be tested for:
- Debian Jessie and Strech
- Ubuntu 16.04 and 18.04

mganeva:
Look to https://nixos.org/patchelf.html and to `cmake RPATH` to fix the dependencies of the built binary

mganeva:
Seel also #472 and try to contact Mika. If he is still working on debian packaging, you might work on it together.

juanmcloaiza:
For future reference:

1. Regarding https://stackoverflow.com/questions/2139124/how-to-make-a-debian-package-depend-on-multiple-versions-of-libboost - Using the OR statement in the Depends section of the deb control file won't change anything, since writing `libb-boost-dev` is already more general than writing `A | B | C`, and we have already tried the former.

Instead, what seems to actually do the trick, is changing the rpath with `patchelf` (example for libboost_program_options; may need `sudo`): 
`$ patchelf --replace-needed libboost_program_options.so.x.y.z libboost_program_options.so /path/to/bornagain_executable`

cmake/bornagain/modules/CPackDebian.cmake
```diff
@@ -24,11 +24,10 @@ set(CPACK_DEBIAN_PACKAGE_NAME "${CPACK_PACKAGE_NAME}")
...
 set(CPACK_DEBIAN_PACKAGE_DEPENDS "libgsl-dev(>=1.15), libboost-date-time${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}, libboost-chrono${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}, libboost-python${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}, libboost-program-options${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}, libboost-iostreams${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}, libboost-regex${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}, libboost-filesystem${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}, libboost-thread${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}, libfftw3-3(>=3.3.1), python3, python3-numpy, python3-matplotlib, libqt5widgets5(>=5.4), libtiffxx5(>=4.0.2)")
```
DmitryYurov: libboost-date-time, chrono, libboost-python have disappeared from the dependencies, but now they are back. It's better to leave regex in the dependencies though.

PR #564, opened 2018-07-16
Print observer for FitObjective

gpospelov:


Core/Fitting/FitObjective.cpp
```diff
@@ -165,6 +162,22 @@ double FitObjective::residual(double a, double b, double weight) const
     return m_chi2_module->residual(a, b, weight);
 }
 
+double FitObjective::evaluate_chi2(const std::vector<double>& residuals,
+                                   const Fit::Parameters& params)
```
DmitryYurov: Possibly, the signature of the form evaluate_chi2(const std::vector<>& sim, const std::vector<>& real, const Fit::Parameters& params) would be better. For the case of Poisson-based error estimate.

gpospelov: Thanks. I didn't look closely into the correctness of "evaluate" related functions. We have to discuss them together with IChiSquared module.

PR #565, opened 2018-07-17
Fix for crash on attempt to edit fit parameter value/limits

DmitryYurov:
Redmine: #2107


GUI/coregui/Models/SessionModelDelegate.cpp
```diff
@@ -76,10 +76,8 @@ QWidget* SessionModelDelegate::createEditor(QWidget* parent, const QStyleOptionV
 {
     auto result = createEditorFromIndex(index, parent);
 
-    auto filter = new TabFromFocusProxy(result);
-    Q_UNUSED(filter);
-
     if (result) {
+        new TabFromFocusProxy(result);
```
gpospelov: I kind of liked previous syntax more. Using Q_UNUSED and variable name serves as a kind of additional hint. I'm kind of uncomfortable with bare "new".

PR #566, opened 2018-07-18
Optimize single 3D Particle creation and population on lattice

anikhalder:
Done by implementing a class Particle3DType which holds a pointer to an instance of the 3D particle to be populated and then by invoking appropriate functions to make copies of the 3D particle and adding them to the RealSpaceModel at the required lattice position. This prevents the re-creation of the 3D particle from scratch (from FormFactor, create material etc.) for the same object that simply needs to be placed at different locations on the lattice.

Even with a very high TotalParticleDensity, the lattice is now populated much faster than before.

Note - this is only for single particle population (population of particle composition, core-shell and distribution yet to be implemented)


GUI/coregui/Views/RealSpaceWidgets/RealSpaceBuilderUtils.h
```diff
@@ -36,12 +37,22 @@ class Particle3DType
...
-    //Particle3DType& operator=(const Particle3DType&) =delete;
-    ~Particle3DType() { //clear();
-                      }
+    Particle3DType& operator=(const Particle3DType& right);
+    ~Particle3DType();
 
-    void clear();
+    void clear3Dparticles();
 
+    const QVector<RealSpace::Particles::Particle*> get3Dparticles() const {return m_3Dparticles; }
```
gpospelov: It's dangerous to give vector of particles  outside. If you need to check size, implement corresponding method

GUI/coregui/Views/RealSpaceWidgets/RealSpaceBuilderUtils.h
```diff
@@ -36,12 +37,22 @@ class Particle3DType
...
+    double getCumulativeAbundance() const {return m_cumulative_abundance; }
```
gpospelov: Please try to avoid "get" prefix in methods.

GUI/coregui/Views/RealSpaceWidgets/RealSpaceBuilderUtils.cpp
```diff
@@ -132,80 +132,56 @@ QVector<double> RealSpaceBuilderUtils::computeCumulativeAbundances(const Session
 }
 
 void RealSpaceBuilderUtils::populateParticlesAtLatticePositions(
-        QVector<QVector<double>> lattice_positions, RealSpaceModel *model,
-        const SessionItem& layoutItem, const SceneGeometry& sceneGeometry,
-        const RealSpaceBuilder *builder3D)
+        QVector<QVector<double>> lattice_positions, const QVector<Particle3DType> &particle3DType_vector,
```
gpospelov: Better to use const reference here.

GUI/coregui/Views/RealSpaceWidgets/RealSpaceBuilderUtils.cpp
```diff
@@ -132,80 +132,56 @@ QVector<double> RealSpaceBuilderUtils::computeCumulativeAbundances(const Session
...
-    QVector<double> cumulative_abundances = computeCumulativeAbundances(layoutItem);
-
     for (QVector<double> position : lattice_positions)
     {
         // for random selection of particles based on their abundances
         double rand_num = (rand()/static_cast<double>(RAND_MAX)); // (between 0 and 1)
         int k = 0;
 
-        for (auto particle : layoutItem.getItems(ParticleLayoutItem::T_PARTICLES))
+        for (auto& particle3DType : particle3DType_vector)
```
gpospelov: Please use const reference

PR #570, opened 2018-07-20
Python callable as FitObjective plot observer

gpospelov:
* Possibility to assign any callable as a drawing observer for FitObjective
* Finally, reproduced functionality of FitCylindersAndPrisms_detailed.py
* ba.FitParameters now silently converted to Python dictionary to have more Pythonic objective functions
* GISASPlotObserverV2 created to plot progress in FitObjective. Old observers will be removed as soon as we get rid from old Python examples.
* SpecularObserver will be updated in the next round.


Examples/python/fitting/ex02_FitBasics/FitCylindersPrisms_detailed_new.py
```diff
@@ -0,0 +1,122 @@
...
+def run_fitting():
+    """
+    Setup simulation and fit
+    """
+
+    real_data = create_real_data()
+
+    fit_objective = ba.FitObjective()
+    fit_objective.addSimulationAndData(get_simulation, real_data, 1.0)
+    fit_objective.initPrint(10)
```
DmitryYurov: This line is doubled.

PR #572, opened 2018-07-24
Optimize single 3D Particle creation and population on lattice

anikhalder:
- Done by implementing a class Particle3DType which holds a pointer to an instance of the 3D particle to be populated and then by invoking appropriate functions to make copies of the 3D particle and adding them to the RealSpaceModel at the required lattice position. This prevents the re-creation of the 3D particle from scratch (from FormFactor, create material etc.) for the same object that simply needs to be placed at different locations on the lattice.

- Revised version with move construct implemented for the class

- Even with a very high TotalParticleDensity, the lattice is now populated much faster than before.

- This is only for single particle population (population of particle composition, core-shell and distribution yet to be implemented)


GUI/coregui/Views/RealSpaceWidgets/RealSpaceBuilderUtils.cpp
```diff
@@ -86,31 +86,31 @@ Particle3DContainer::~Particle3DContainer()
 
 void Particle3DContainer::clearContainer()
 {
-    for (auto it = m_Particle3DContainer.begin(); it != m_Particle3DContainer.end(); ++it)
+    for (auto it = m_container.begin(); it != m_container.end(); ++it)
```
gpospelov: At some point you might want to replace it with range based loop

GUI/coregui/Views/RealSpaceWidgets/RealSpaceBuilderUtils.cpp
```diff
@@ -55,35 +55,59 @@ namespace
...
-    for (int i = 0; i < m_container.size(); ++i)
-        m_container[i] = new RealSpace::Particles::Particle(*p3D.m_container[i]);
+    m_container.resize(other.m_container.size());
+    for (size_t i = 0; i < m_container.size(); ++i)
+        m_container[i] = new RealSpace::Particles::Particle(*other.m_container[i]);
 }
 
-Particle3DContainer &Particle3DContainer::operator=(const Particle3DContainer &right)
+// copy assignment
+Particle3DContainer &Particle3DContainer::operator=(const Particle3DContainer &rhs)
```
gpospelov: You have inconsistent  usage of "space" in type qualifiers. We are using `const ParticleContainer& rhs`. Please note the space between name and reference.

Tests/UnitTests/GUI/TestRealSpaceBuilderUtils.h
```diff
@@ -77,34 +77,35 @@ TEST_F(TestRealSpaceBuilderUtils, test_Particle3DContainer)
     p1.setCumulativeAbundance(1);
     p1.setParticleType(Constants::ParticleType);
 
-    EXPECT_EQ(p1.containerSize(), 1);
+    EXPECT_EQ(p1.containerSize(), static_cast<size_t>(1));
```
gpospelov: Use '1u' instead of static_cast

PR #581, opened 2018-07-26
GUI3D ParticleComposition population from Particle3DContainer

anikhalder:
- A vector of bool has been added to Particle3DContainer class in order to keep track which particle in the container should have a blend (transparency). This is particularly useful for ParticleCoreShell.


Tests/UnitTests/GUI/TestRealSpaceBuilderUtils.h
```diff
@@ -77,10 +77,12 @@ TEST_F(TestRealSpaceBuilderUtils, test_Particle3DContainer)
     p1.addParticle(cylinder3D.release());
     p1.setCumulativeAbundance(1);
     p1.setParticleType(Constants::ParticleType);
+    p1.fillContainerParticlesBlend(false);
```
gpospelov: I think you do not need the method fillContainerParticlesBlend(). It's too complicated to keep two different calls synchronized. You can add second bool parameter to method  Particle3DContainer::addParticle

Tests/UnitTests/GUI/TestRealSpaceBuilderUtils.h
```diff
@@ -146,6 +153,7 @@ TEST_F(TestRealSpaceBuilderUtils, test_Particle3DContainer)
     EXPECT_EQ(p6.containerSize(), static_cast<size_t>(1));
```
gpospelov: Use "1u" instead of static_cast

PR #586, opened 2018-07-27
Fix issue #2125: Calculation of maximum reciprocal lattice points to consider for 2d lattice is wrong

waltervh:


Core/Aggregate/InterferenceFunction2DLattice.cpp
```diff
@@ -190,18 +190,24 @@ InterferenceFunction2DLattice::transformToPrincipalAxes(double qx, double qy, do
...
-    double qy_rot = -qx*std::sin(xi) + qy*std::cos(xi);
-
-    int qa_int = static_cast<int>(std::lround(a*qx_rot / M_TWOPI));
-    int qb_int = static_cast<int>(
-                     std::lround(b*(qx*std::cos(alpha) + qy*std::sin(alpha)) / M_TWOPI));
+    // first rotate the input to the system of m_sbase:
+    double qx_rot = qx * std::cos(xi) + qy * std::sin(xi);
+    double qy_rot = -qx * std::sin(xi) + qy * std::cos(xi);
+
+    // find the reciprocal lattice coordinates of (qx_rot, qy_rot):
```
gpospelov: I think removal of 2pi factor is wrong

waltervh: yes, indeed! thanks for pointing that out!

gpospelov:
Impressive, thanks!

PR #595, opened 2018-07-31
Provide possibility to have the reference point of full spheres at their center instead of at the bottom (issue #2132)

waltervh:


Core/HardParticle/FormFactorFullSphere.cpp
```diff
@@ -23,8 +23,8 @@
 
 //! Constructor of a full sphere.
 //! @param radius: radius of the sphere in nanometers
```
DmitryYurov: Would be nice to extend the description

PR #597, opened 2018-08-01
Redmine issue 2134: create a shortcut for translating numpy array in OutputData

DmitryYurov:


Core/Instrument/ArrayUtils.h
```diff
@@ -15,126 +15,35 @@
...
-
-//! Creates OutputData from 1D vector.
-//! @param vec: std::vector<double>
-//! @return std::unique_ptr<OutputData<double>>
-template<class T> decltype(auto) createData1D(const T& vec);
-
-//! Creates 1D vector from OutputData.
-//! @param vec: OutputData<double>
-//! @return vector<double>
-template<class T> decltype(auto) createVector1D(const T& data);
```
gpospelov: If I'm not mistaken, you completely removed functions createVector1D basing on the fact that they are not used? This pair (data to vector, vector to data) was created and tested to replace in the future functions from IntensityDataFunctions namespace. As you know, our FFT machinery requires translation in both directions. For the moment this is done with the help of functions like `create2DArrayfromOutputData`. They are not correct. The idea was to replace them.

Core/Instrument/ArrayUtils.h
```diff
@@ -15,126 +15,35 @@
...
-
-//! Creates OutputData from 2D vector.
-//! @param vec: std::vector<std::vector<double>>
-//! @return std::unique_ptr<OutputData<double>>
-template<class T> decltype(auto) createData2D(const T& vec);
```
gpospelov: My idea of using templates here was to create universal way to convert not only doubles, but for example integer types.

DmitryYurov: Let us see what `git grep -e "OutputData<" --and --not -e "OutputData<double>" --and --not -e "OutputData<T>" -- "*.h" "*.cpp"` shows on current `develop` of `scgmlz/BornAgain` (I have also removed SWIG-generated files manually):

```
Core/Instrument/ArrayUtils.h:    std::unique_ptr<OutputData<value_type>> result(new OutputData<value_type>());
Core/Instrument/ArrayUtils.h:    std::unique_ptr<OutputData<value_type>> result(new OutputData<value_type>());
Core/Instrument/DetectorMask.h:    const OutputData<bool>* getMaskData() const { return &m_mask_data; }
Core/Instrument/DetectorMask.h:    OutputData<bool> m_mask_data;
Core/Instrument/Histogram1D.cpp:    OutputData<CumulativeValue>::const_iterator it_origin = m_data.begin();
Core/Instrument/Histogram1D.cpp:    OutputData<CumulativeValue>::iterator it_result = result->m_data.begin();
Core/Instrument/Histogram2D.cpp:    OutputData<CumulativeValue>::const_iterator it_origin = m_data.begin();
Core/Instrument/Histogram2D.cpp:    OutputData<CumulativeValue>::iterator it_result = result->m_data.begin();
Core/Instrument/IDetectorResolution.h:    void applyDetectorResolutionPol(OutputData<Eigen::Matrix2d>* p_matrix_intensity) const;
Core/Instrument/IHistogram.cpp:    OutputData<CumulativeValue>::const_iterator it =
Core/Instrument/IHistogram.cpp:    OutputData<CumulativeValue>::const_iterator it =
Core/Instrument/IHistogram.cpp:    OutputData<CumulativeValue>::const_iterator it =
Core/Instrument/IHistogram.h:    const OutputData<CumulativeValue>& getData() const { return m_data; }
Core/Instrument/IHistogram.h:    OutputData<CumulativeValue>& getData() { return m_data; }
Core/Instrument/IHistogram.h:    OutputData<CumulativeValue> m_data;
Core/Instrument/IIntensityNormalizer.cpp:    OutputData<double >* normalized_data = data.clone();
Core/Instrument/OutputData.cpp:double OutputData<CumulativeValue>::getValue(size_t index) const
Core/Instrument/OutputData.h:    template <class U> void copyShapeFrom(const OutputData<U>& other);
Core/Instrument/OutputData.h:    template <class U> bool hasSameDimensions(const OutputData<U>& right) const;
Core/Instrument/OutputData.h:    template <class U> bool hasSameShape(const OutputData<U>& right) const;
Tests/UnitTests/Core/DataStructure/OutputDataTest.h:    OutputData<int> int_data_0d;
Tests/UnitTests/Core/DataStructure/OutputDataTest.h:    OutputData<float> fl_data_1d;
Tests/UnitTests/Core/DataStructure/OutputDataTest.h:    OutputData<Eigen::Matrix2d> matrix_data_2d;
Tests/UnitTests/Core/DataStructure/OutputDataTest.h:    OutputData<Eigen::Matrix2d> mdata1;
Tests/UnitTests/Core/DataStructure/OutputDataTest.h:    OutputData<Eigen::Matrix2d> mdata2;
Tests/UnitTests/Core/DataStructure/OutputDataTest.h:    OutputData<Eigen::Matrix2d> mdata;
Tests/UnitTests/Core/DataStructure/OutputDataTest.h:    OutputData<bool> data_bool;
Tests/UnitTests/Core/DataStructure/OutputDataTest.h:    OutputData<bool> data_double;
Tests/UnitTests/Core/Detector/SphericalDetectorTest.h:    const OutputData<bool>* mask = detector.detectorMask()->getMaskData();
```

Apart from tests, `OutputData.h` and `ArrayUtils.h`, it counts to 13 use cases, majority of which are from `Histogram` classes (i.e. are related to stalled entities). None of them are due to `OutputData<int>`, by the way.

Thus I consider using templates as an obvious premature optimization in this case. Especially taking into account, that the functions can be hardly used with anything except `std::vector` or `std::vector<std::vector>`.

Core/Instrument/ArrayUtils.h
DmitryYurov: Yes, I removed them, since they are completely useless: all the required functionality was already implemented in `OutputData::getRawDataVector()`. If you are not happy with `IntensityDataFunctions::create2DArrayfromOutputData`, it is much more reasonable to use the mentioned method of `OutputData` and then to manipulate the vector obtained.

If FFT is the only use case for the functions I have deleted, consider making some local changes to `IntensityDataFunctions` (e.g. renaming and slight modification of `create2DArrayfromOutputData`) instead of extracting a new set of acquisition functions.

Core/Instrument/ArrayUtils.cpp
```diff
@@ -43,4 +46,51 @@ PyObject* ArrayUtils::createNumpyArray(const std::vector<double>& data)
...
+    size_t ncols(0);
+    if(nrows) ncols = data[0].size();
+    for(size_t row=0; row<nrows; row++)
+        if(data[row].size() != ncols)
+            throw std::runtime_error("Util::getShape() -> Error. "
+                                     "Number of elements is different from row to row.");
+    return std::make_pair(nrows, ncols);
+}
+
+OutputData<double>* ArrayUtils::createData1D(const std::vector<double>& vec)
```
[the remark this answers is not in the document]
DmitryYurov: 1. Your proposal will require further expansion of swig-configuration files (`*.i`), which I would like to avoid (since they are quite messy, and swig behaviour sometimes gets me nuts). Not a strong argument, but your arguments are even weaker (for other comments see below).

2. Returning naked pointer instead of `std::unique_ptr` makes not so much of a difference. If you look at the code attentively, every use of `createData1D` and of `createData2D` is left fully compatible with returning `std::unique_ptr`. Thus switch back to unique pointers will come at no cost.

3. About temporality: these functions will be used at least one release, may be longer. When considering new functionality, places in the code with a comparable lifetime are quite numerous.

PR #600, opened 2018-08-02
New formfactors for microgels

mganeva:
Ornstein-Zernike and Debye-Bueche formfactors, unit test and updated Python wrappers.


waltervh:
In the future, please start right away from the develop branch. It still looks very ok on github, but the automatic email I got was filled with commits between master and develop ;-)

PR #602, opened 2018-08-08
GUI3D Remove offset of particles during 3D particle construction

anikhalder:
- The BaseShapes have been refactored such that the bottom of the objects lie in the z=0 plane during object creation and there remains no need for separately specifying the offset of the 3D particles (as done previously).

- This also fixes the error in the rotation of the 3D particle due to the non-zero offset parameter which was being specified previously.

### DIFF
--- a/GUI/ba3d/ba3d/model/particles.cpp
+++ b/GUI/ba3d/ba3d/model/particles.cpp
@@ -86,7 +86,6 @@ AnisoPyramid::AnisoPyramid(float L, float W, float H, float alpha)
     isNull = (L <= 0 || W <= 0  || H <= 0 || alpha <= 0);
     turn = Vector3D(0,0,45*pi/180.0f);
     scale  = Vector3D(L*sqrt2f, W*sqrt2f, H);
-//    offset = Vector3D(0, 0, H/2); // (PREVIOUSLY)
     offset = Vector3D(0, 0, 0);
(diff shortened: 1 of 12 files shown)


GUI/ba3d/ba3d/model/particles.cpp
```diff
@@ -64,189 +83,191 @@ void Particle::addTranslation(Vector3D translate_)
...
 // see ~/BornAgain/GUI/ba3d/ba3d/model/geometry/ for BaseShape construction
+// e.g. column.cpp, sphere.cpp etc.
+
+/* PREVIOUSLY, the base shapes were constructed symmetric about the xy plane, centered at the
+origin. E.g. for column base shape (see column.cpp), the xy-plane went through the center of the
+shape (object extension: -H/2 to +H/2 in z). When a 3D particle was created (see the following code)
+using this column base shape and scaled to proper dimensions, the offset  was kept at z=H/2 to bring
+the bottom of the particle to z=0 plane. NOW, the base shapes are all created such that the bottom
+of the particle is already in the z=0 plane and there is no need for specifying the bottom offset -
+in other words, offset = Vector3D(0,0,0) */
```
waltervh: I think you forgot to clean this comment up (remove PREVIOUSLY etc.) Never mind though. Will be done in future commits.

PR #607, opened 2018-08-10
New assistant to import csv (char-separated-value) files.

juanmcloaiza:
On data import exceptions, a GUI is presented to the user to select the region of the file to import and the character with which columns are separated. A preview of the datafile is made available to the user in the form of a QTableWidget.


juanmcloaiza:
Windows builds not passing; investigating further...

PR #614, opened 2018-09-02
Bug-fix for showing wrong results in GUI on fitting finish

DmitryYurov:


DmitryYurov:
Have forgotten, that apps.fz-juelich is down. Closing for now

PR #623, opened 2018-09-13
Data imported without creating an intermediate file + some bugs corrections.

juanmcloaiza:
The windows build was fixed by `#include "../Core/InputOutput/DataFormatUtils.cpp"`. This allows for not only the declarations (in the .h) but also the implementations (in the .cpp) of the classes/functions there to be accessible; in particular `std::vector<double> DataFormatUtils::parse_doubles(const std::string& str)`. There are two ways (that I'm aware of) of making these classes and functions accessible to the `CsvImportAssistant`:

1. The neat way: at cmake level
2. The patchy way: adding `#include relative/path/to/file/containing/implementations.cpp` where needed (as I did here)

Mixing the Core functionality and the GUI functionality at cmake level seemed to entangle two worlds that are separated by design (as far as I can tell).

Given that `DataFormatUtils::parse_doubles(const std::string& str)` returns a vector of doubles instead of a single double, and inside it some string to stream conversion takes place, its use at firs seems a bit of an overkill for parsing a single double from a string; thus, parsing single doubles using the `atof`and `strtod` was also tried, however, the parsing was not being done correctly. Additionally, these functions have not been used so far in BornAgain (as far as `grep` can tell), thus, the most reasonable choice is to attach to the already implemented functionality (`DataFormatUtils::parse_doubles(const std::string& str)`) and use it in the `CsvImportAssistant`, even if it is an overkill for our needs. On top of that, `DataFormatUtils::parse_doubles(const std::string& str)` produces trusted results.

### DIFF
--- a/Core/InputOutput/OutputDataReadFactory.cpp
+++ b/Core/InputOutput/OutputDataReadFactory.cpp
@@ -28,8 +28,6 @@ IOutputDataReadStrategy* OutputDataReadFactory::getReadStrategy(const std::strin
     IOutputDataReadStrategy* result(nullptr);
     if(DataFormatUtils::isIntFile(file_name))
         result = new OutputDataReadINTStrategy();
-    else if(DataFormatUtils::isTxtFile(file_name))
-        result = new OutputDataReadNumpyTXTStrategy();
 #ifdef BORNAGAIN_TIFF_SUPPORT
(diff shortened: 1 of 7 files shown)


waltervh:
I will temporarily close this until you have cleaned the commit history.

PR #625, opened 2018-09-13
Dropping OutputDataReadNumpyTXTStrategy in favor of CsvImportAssistant

juanmcloaiza:


GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant.cpp
```diff
@@ -220,34 +229,75 @@ void CsvImportAssistant::onRejectButton(){
...
+    vector<string> StringVector;
 
     //save the values of the array
-    for(int i = 0; i < nRows; i++){
-        B.clear();
-        for(int j = 0; j < nCols; j++){
-            B.push_back( m_tableWidget->item(i,j) != nullptr ? m_tableWidget->item(i,j)->text().toStdString() : "");
+    std::size_t nDataCols = 0;
+    std::size_t nDataRows = 0;
+    for(int i = 0; i < nTableRows; i++){
```
DmitryYurov: The performance of this loop could be better. Could you please check if it works well on large data arrays (e.g. 1000X1000)? It is not the only place where similar logic (with pushing back to string arrays) is used

GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant.cpp
```diff
@@ -220,34 +229,75 @@ void CsvImportAssistant::onRejectButton(){
...
 
-    setFilepath(filepath()+"-ba-imported.txt");
-    ofstream output_file(filepath().toStdString());
-    for(int i = 0; i < nRows; i++){
-        for(int j = 0; j < nCols - 1; j++){
-            std::string a = A[unsigned(i)][unsigned(j)];
-            output_file << a << " ";
+    std::unique_ptr<OutputData<double>> result, result1d; //= new OutputData<double>;
+    result.reset(new OutputData<double>());
+    result1d.reset(new OutputData<double>());
```
DmitryYurov: It is a little bit strange to have result and result1d at once. Only one of them is finally used, right? Consider rewriting this piece of code to exclude unnecessary operations.

GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant.cpp
```diff
@@ -220,34 +229,75 @@ void CsvImportAssistant::onRejectButton(){
...
 {
     using namespace std;
 
-    int nRows = m_tableWidget->rowCount();
-    int nCols = m_tableWidget->columnCount();
-    vector<vector<string>> A;
-    vector<string> B;
+    int nTableRows = m_tableWidget->rowCount();
+    int nTableCols = m_tableWidget->columnCount();
+    vector<vector<string>> StringVectorVector;
```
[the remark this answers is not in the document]
juanmcloaiza: 1. The link you provide (pps.jcns.fz-juelich.de/dok/sc/intern:coding_rules) is broken, but I know what you mean.
2. "temporary_storage" doesn't give any additional information either :)

GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant.cpp
juanmcloaiza: Could you provide me with real test cases? That way I could provide the group with a relevant profile of the performance of the importer, thanks.

Here's a comparison of push_back() between `vector` and `dequeue` containers, to have an idea of the kind of numbers you can expect:

https://www.codeproject.com/Articles/5425/An-In-Depth-Study-of-the-STL-Deque-Container

GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant.cpp
juanmcloaiza: Sure, let me take a look... The aim was to use the code that was already written.

GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant.cpp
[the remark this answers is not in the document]
waltervh: The correct link is:
http://apps.jcns.fz-juelich.de/doku/sc/intern:coding_rules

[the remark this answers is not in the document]
DmitryYurov: Sorry, lost "u" in "doku" indeed. Concerning "temporary_storage", I do not insist on this name in particular. If you think some other name fits better, you are (of course) free to use it. But "StringVectorVector" just repeats the type. Its purpose is not obvious from the naming, which is undesirable.

GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant.cpp
DmitryYurov: Probably, I expressed my thought vaguely. push_back to string arrays is mainly mentioned to identify similar places, though push_back itself has nearly constant complexity. The performance could be improved by the following actions:
1. Reserving space for elements in advance of actual push_back
2. Avoiding usage of the second intermediate array (StringVector), pushing directly in StringVectorVector
3. Using move instead of copying (there is a special push_back overload for this case)

All these improvements make sense only if you can notice lags on rather big data arrays. You can consider manually generated txt-file with 1000x1000 entries. It can be done with e.g. gawk or just bash. 

Of course the main use case for the loader is loading 1D data, so this improvement is not so principal for the pull-request.

juanmcloaiza: Exactly, that was my guess: the extra performance gained would be unnecessary for our use case. Unless you are really planing on loading files that large, I wouldn't focus as much on performance as on readability.

PR #630, opened 2018-09-17
Fix for loading old projects

DmitryYurov:
This issue is solved seemingly easy by skipping update to the instrument if no data item in RealDataItem is available. However, it makes the code more fragile, since now we rely on the integrity of saved data. Still it is required only for old projects.

I would be very grateful for the callback (if this fix is ok or should I find a better way).


waltervh:
Please close this issue for a while. We are discussing here what would be the best option. I think your fix might be the best solution, but there are some other options to consider.

DmitryYurov:
Ok then, closing. One more notion: this problem will be solved automatically after finishing transition to universal DataItem and view proxies.

PR #638, opened 2018-09-27
Fixed rare crash of fitting job

gpospelov:
+ Bug 2169 - 	Application crash on fitting job
+ Bug 2172 - 	Fit flow widget doesn't update after fit re-launch

### DIFF
--- a/Core/Fitting/FitSuite.cpp
+++ b/Core/Fitting/FitSuite.cpp
@@ -156,6 +156,12 @@ FitSuiteStrategies* FitSuite::fitStrategies()
     return m_impl->fitStrategies();
 }
 
+bool FitSuite::isFirstIteration() const
+{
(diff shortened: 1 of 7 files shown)


DmitryYurov:
1. Long fitting runs are now slowed down on last iterations. Try fitting SpecularSimulation with genetic algorithm (30 iterations, 30 specimens, plot update slider to "1" position), it will slow down after 600 hundred iterations or so. It was keeping the same fitting pace before the changes. 

2. Logged results show that the last iteration is repeated two times. It was so also before the changes, but it is the possible source of the previously observed crash. Troubles because of access to `m_block_plots_update` before locking mutex are quite unprobable. I would try make it atomic and see what changes.

double iteration in the end

gpospelov:
Ok, thanks. Will try to find out the reason of slowing.

PR #650, opened 2018-10-05
Csv import assistant added

juanmcloaiza:
## Tests made:
 - Import of a 150x150 number array  
  - Automatically guess separator without crashing
  - "Separator" field updates table as expected.
  - "Column" field updates table as expected.
  - "From row" field updates table as expected.
 - Does not crash when reading empty file.
- Does not crash when reading corrupted / binary file (shows rubbish on the table)
- *Read* 1200 x 1200 string array in 3 seconds
- *Import* 1200 x 1200 number array in 6 seconds (bad implementation, must be revisited)
- When hitting import an error message is shown if the data is not "good" (does not crash).
- Context menu works as expected (updates headers, updates table and updates controls in the bottom of the dialog).
- Import 1D works in two possible ways: a) setting 1 column as coordinate and 1 as intensity and then hitting the import button, or 2) setting one column as "intensity bins".

## Desired:
- Do not make columns/rows disappear from the table but instead grey them out.
- Let the user choose between importing 1d or 2d BEFORE opening the dialog
- If the 2d array can be imported automatically (e.g. with the machinery for importing numpy arrays saved as txt), do it without opening the dialog.

### DIFF
--- a/GUI/coregui/Views/ImportDataWidgets/ImportDataUtils.cpp
+++ b/GUI/coregui/Views/ImportDataWidgets/ImportDataUtils.cpp
@@ -21,6 +21,7 @@
 #include "IntensityDataIOFactory.h"
 #include "IntensityDataItem.h"
 #include "RealDataItem.h"
+#include "CsvImportAssistant.h"
 #include "projectmanager.h"
 #include <QFileDialog>
 #include <QFileInfo>
@@ -28,7 +29,7 @@
 
 namespace
 {
-const QString filter_string = "Intensity File (*.int *.gz *.tif *.tiff *.txt);;"
+const QString filter_string = "Intensity File (*.int *.gz *.tif *.tiff *.txt *.csv);;"
                               "Other (*)";
 
 int getRank(const RealDataItem& item)
@@ -44,8 +45,14 @@ int getRank(const InstrumentItem& item) {
 std::unique_ptr<OutputData<double>> ImportDataUtils::ImportData(QString& baseNameOfLoadedFile)
 {
     QString dirname = AppSvc::projectManager()->userImportDir();
-    QString fileName = QFileDialog::getOpenFileName(0, QStringLiteral("Open Intensity File"),
+    QString fileName = QFileDialog::getOpenFileName(Q_NULLPTR, QStringLiteral("Open Intensity File"),
                                                     dirname, filter_string);
+
+    std::unique_ptr<OutputData<double>> result;
+
+    QString newImportDir = GUIHelpers::fileDir(fileName);
+    if (newImportDir != dirname)
+        AppSvc::projectManager()->setImportDir(newImportDir);
 
     if (fileName.isEmpty())
         return nullptr;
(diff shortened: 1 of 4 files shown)


GUI/coregui/Views/ImportDataWidgets/ImportDataUtils.cpp
```diff
@@ -44,35 +45,58 @@ int getRank(const InstrumentItem& item) {
...
+                    IntensityDataIOFactory::readOutputData(fileName.toStdString()));
         result = CreateSimplifiedOutputData(*data.get());
-    } catch (std::exception& ex) {
-        QString message = QString("Error while trying to read file\n\n'%1'\n\n%2")
-                              .arg(fileName)
-                              .arg(QString::fromStdString(std::string(ex.what())));
-        QMessageBox::warning(0, "IO Problem", message);
-    }
 
+    } catch(std::exception& e)
```
DmitryYurov: You could replace it with just `} catch(std::exception&) {` to avoid generating warnings about unused exception parameter.

GUI/coregui/Views/ImportDataWidgets/ImportDataUtils.cpp
```diff
@@ -44,35 +45,58 @@ int getRank(const InstrumentItem& item) {
...
+        if(res == assistant.Accepted){
+            result = assistant.getData();
+            return true;
+        }
+    }catch(std::exception& e){
+        QString message = QString("Unable to read file:\n\n'%1'\n\n%2\n\nCheck that the file exists and it is not being used by other program.\n\n")
+                .arg(fileName)
+                .arg(QString::fromStdString(std::string(e.what())));
+        QMessageBox::warning(nullptr, "IO Problem", message);
+        return false;
```
DmitryYurov: Statement is unnecessary. The next one will work anyway. I would remove it.

GUI/coregui/Views/ImportDataWidgets/ImportDataUtils.h
```diff
@@ -31,6 +31,8 @@ namespace ImportDataUtils
 {
 
 BA_CORE_API_ std::unique_ptr<OutputData<double>> ImportData(QString& baseNameOfLoadedFile);
+BA_CORE_API_ bool UseImportAssistant(QString& fileName, std::unique_ptr<OutputData<double>>& result);
```
DmitryYurov: I would change the signature to

```
std::unique_ptr<OutputData<double>> UseImportAssistant(const QString& fileName);
```
Failure can be reported by empty `unique_ptr`.

GUI/coregui/Views/ImportDataWidgets/ImportDataUtils.cpp
```diff
@@ -44,35 +45,58 @@ int getRank(const InstrumentItem& item) {
...
-        QString message = QString("Error while trying to read file\n\n'%1'\n\n%2")
-                              .arg(fileName)
-                              .arg(QString::fromStdString(std::string(ex.what())));
-        QMessageBox::warning(0, "IO Problem", message);
-    }
 
+    } catch(std::exception& e)
+            //Try to import data using the GUI importer
+    {
+        std::unique_ptr<OutputData<double>> data;
```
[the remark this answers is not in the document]
juanmcloaiza: Yes, you're right, the user must select in advance whether to load 1D data or 2D data (to be implemented for the next Pull Request).

PR #656, opened 2018-10-08
1. Grey out data not to be imported; 2. Disable 2d import from the assistant

juanmcloaiza:


GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant.cpp
```diff
@@ -354,6 +351,34 @@ void CsvImportAssistant::set_table_data(std::vector<std::vector<std::string>> da
...
+        }
+
+    //Return if there are no columns to grey out
+    if( m_coordinateCol + m_intensityCol + m_singleCol < 1 )
+        return;
+
+    //Grey out columns
+    for(int i = 0; i < nRows; i++)
+        for(int j = 0; j < nCols; j++)
+            if(j+1 != int(m_coordinateCol) &&
```
gpospelov: It's almost impossible to encode what is going on here. Please consider simplifying of if statement with some helper functions.

PR #657, opened 2018-10-09
Redmine issue 2177: PointwiseAxisItem and associated save/load mashinery

DmitryYurov:
1. Inclination angle axis in `SpecularInstrumentItem` is now handled through an axis group, which holds both `BasicAxisItem` and `PointwiseAxisItem`. It enables simpler handling of non-xml data.

2. `ImportDataInfo` class stays a little bit of the path of the pull-request. It was introduced to let Juan to proceed with loading data from ASCII files. It is assumed, that this class will replace `std::unique_ptr<OutputData<double>>` in `ImportDataUtils::Import1dData`.

3. `PointwiseAxisItem::P_FILE_NAME` is bound to `InstrumentItem::P_IDENTIFIER`, which helps to keep file names unique for each instrument

### DIFF
--- a/GUI/coregui/Models/ApplicationModels.cpp
+++ b/GUI/coregui/Models/ApplicationModels.cpp
@@ -231,8 +231,9 @@ QList<SessionModel*> ApplicationModels::modelList()
 
 QVector<SessionItem*> ApplicationModels::nonXMLData() const
 {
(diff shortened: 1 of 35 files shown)


DmitryYurov:
Thanks!

PR #662, opened 2018-10-12
RealSpaceWidget in a dock

gpospelov:
+ Updated code of FancyMainWindow from qt-creator source (it gave us dynamic toolbars in docks)
+ Created dedicated dock for 3D view
+ The old way of calling RealSpaceWidget is still working in parallel 
+ Removed custom tool bar from PyScriptWidget to have all docks uniform

Remaining things:
+ Tune default sizes of RealSpaceWiddget and PyScriptWidget  on corresponding dock hide/show.
+ Provide a way to drag RealSpaceWidget across graphical scene without occasional embedding in nearest docking area.


DmitryYurov:
Looks pretty. Several comments though:

1. User can see now two buttons for essentially the same functionality: `Real Space Viewer` and `3D Viewer`. It is just that one of them opens sample view in a separate window. It is ok, if you will remove the old button in the future.

2. It is hard to undock `3D Viewer` from bottom panel. I need to click two times on the panel head to get widget edge. After that I can undock & drag it. Wouldn't it be nice, if I could drag the widget just by its head?

3. After docking the widget to the left-side panel, the only way to change its position is to restore default layout from docks layout menu.

4. The least important: the first time I open the widget with the `3D Viewer` button, I can see only its head. After resizing it memorizes its own size, thus it is not a big problem. Still it would be better, if it would have non-zero default size.

PR #668, opened 2018-10-17
 Bug 2183 fixed: Adding extension when saving python scripts 

juanmcloaiza:


GUI/coregui/Views/SimulationWidgets/PythonScriptWidget.cpp
```diff
@@ -104,19 +104,28 @@ void PythonScriptWidget::generatePythonScript(const MultiLayerItem *sampleItem,
...
+
 void PythonScriptWidget::onExportToFileButton()
 {
     QString dirname(m_outputDir);
     if (dirname.isEmpty())
         dirname = QStandardPaths::writableLocation(QStandardPaths::HomeLocation);
 
-    QString file_name = QFileDialog::getSaveFileName(
-        this, "Select file", dirname, "Python scipts (*.py)", 0, QFileDialog::DontResolveSymlinks);
+    QString filters("Python scripts (*.py);;All files (*.*)");
```
DmitryYurov: I would suggest to remove "All files". This option saves file to unknown place anyway (or doesn't save at all?) if no extension was specified. Moreover, user will hardly ever use it.

GUI/coregui/Views/SimulationWidgets/PythonScriptWidget.cpp
```diff
@@ -104,19 +104,28 @@ void PythonScriptWidget::generatePythonScript(const MultiLayerItem *sampleItem,
...
+    QString defaultFilter("Python scripts (*.py)");
+    QString defaultName = dirname + QString("/untitled.py");
```
DmitryYurov:
```suggestion
    QString defaultName = dirname + QString("/untitled");
```

It will keep dialog in line with all the other save dialogs of BornAgain.

PR #682, opened 2018-11-06
Remove unused STGZ package generator

jwuttke:
It generated a Unix install shell script (*.sh) that has never
been used.

Also, remove the setting of CPACK_SOURCE_GENERATOR, which duplicates
CPACK_GENERATOR set a few lines above.

This resolves #2216.


cmake/bornagain/modules/BornAgainCPack.cmake
```diff
@@ -34,12 +34,11 @@ elseif(APPLE)
 elseif(UNIX AND BUILD_DEBIAN) # one can build debian package only on UNIX system
     include(CPackDebian)
 else()
-  set(CPACK_GENERATOR "STGZ;TGZ")
+  set(CPACK_GENERATOR "TGZ")
 endif()
 
 
 # Generating the source package
-set(CPACK_SOURCE_GENERATOR "TGZ")
```
DmitryYurov: This line does not duplicate `CPACK_GENERATOR`. After it has been removed, attempt to generate tarball (`make package_source`) leads to generating four different source archives: `tar.bz2`, `tar.gz` (which is the desired one), `tar.xz` and `tar.Z`.

waltervh:
Closed because breaks tarball generation

PR #698, opened 2018-11-19
CsvImportAssistant: Add multiplier functionality + do some refactoring.

juanmcloaiza:


GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant/CsvDataColumn.h
```diff
@@ -0,0 +1,95 @@
...
+
+    // Getters:
+    int columnNumber() const { return m_colNum; }
+    double multiplier() const { return m_multiplier; }
+    csv::DataColumn values() const { return m_values; }
+
+    // Setters:
+    void setColNum(int const colNum) { m_colNum = colNum; }
+    void setMultiplier(double const multiplier) { m_multiplier = multiplier; }
+    void setValues(csv::DataColumn const values) { m_values = values; }
```
DmitryYurov: This method is later used in a context, which allows `move` semantics (namely, in `CsvImportTable::setColumnAs` method). For this reason I would implement it as

```
setValues(csv::DataColumn&& values) {m_values = std::move(values);}
```
or something similar. This will improve performance by avoiding extensive copying. 
This comment holds true for many places in the code.
You can learn about move semantics from "Effective modern C++" by Scott Meyers, chapter 5.

GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant/CsvDataColumn.h
```diff
@@ -0,0 +1,95 @@
...
+    void resetColumn(int colNum = -1, double multiplier = 1., csv::DataColumn values = {})
+    {
+        m_colNum = colNum;
+        m_multiplier = multiplier;
+        m_values = values;
+    }
+    CsvIntensityColumn& operator=(const CsvIntensityColumn colToRep)
```
DmitryYurov: Placing implementations of methods (instead of just declarations) in a header is generally a bad idea (except for templates and short inlined getters/setters). Implementation in a class declaration is an implicit request for inlining + it usually requires additional header inclusions, which slows down compilation process. Please place implementations into cpp (together with associated includes).

GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant/CsvDataColumn.h
```diff
@@ -0,0 +1,44 @@
...
+const QStringList HeaderLabels{"Intensity", "theta", "q"};
+const QStringList UnitsLabels{"default", "bin", "rad", "deg", "mm", "1/nm"};
+
+class CsvDataColumn
+{
+public:
+    // Constructors:
+    CsvDataColumn() : m_units(AxesUnits::NBINS), m_multiplier(1.0), m_values({}) {}
+
+    CsvDataColumn(CsvDataColumn& toCopy)
```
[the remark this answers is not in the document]
waltervh: Non-const arguments of copy constructors prohibit copying a const object. If they really change the copied object, they display uncommon copy semantics, which should be clearly documented then. The infamous std::auto_ptr was a clear example of a class with weird copy semantics.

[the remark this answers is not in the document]
DmitryYurov: Exactly, very good explanation

GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant/CsvDataColumn.h
waltervh: I would not make this type of methods use move semantics because it can possibly improve performance. It is better to measure before optimizing. On the other hand, nothing prevents you from using
`m_values = std::move(values)`
in the body of the method, to prevent copying twice.

GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant/CsvDataColumn.h
gpospelov: I would close the request and then reopen it when everything is fixed to make life of Travis a bit easier.

GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant/CsvDataColumn.h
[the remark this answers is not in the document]
waltervh: Just as an extra remark, copy constructing by non-const reference also removes the possibility to copy from rvalues.

GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant/CsvDataColumn.h
DmitryYurov: Minor clarification: passing by value and moving will provide the required functionality here without restricting method only to r-values:

```
void setValues(csv::DataColumn values) {m_values = std::move(values);}
```

juanmcloaiza: I wonder if this post gives some ideas... https://zpjiang.me/2018/08/01/Copy-Elision-in-C/

I have gone through the code and there are so many places in Bornagain where one could make `something = std::move(somethingElse)`. However, `std::move` is used only 141 times (in code not belonging to the tests or to ThirdParty), of which 17 (i.e. more than 10% were just introduced now).

waltervh: BornAgain was developed in a pre-C++11 era. The presence of suboptimal code in a code base is not a good argument to do the same.

PR #699, opened 2018-11-19
Clean fit examples from old FitSuite

gpospelov:


Examples/python/fitting/ex01_BasicExamples/minimizer_settings.py
```diff
@@ -53,20 +53,36 @@ def get_simulation(params):
     return simulation
 
 
+def create_real_data():
+    """
+    Generating "real" data by adding noise to the simulated data.
+    This function has been used once to generate refdata_fitcylinderprisms.int
+    located in same directory.
+    """
```
waltervh: This file is not generated anymore right? And is there any noise added?

PR #700, opened 2018-11-20
Redmine issue 2222: better representation/editing in tuning tree view

DmitryYurov:
Features:

1. Fancy exponential/floating format in tree view (both for model and spin box)
2. Free edit of parameter value (validation only on focus loss or pressing enter)
3. The same increment/decrement step in slider and spin box

### DIFF
--- a/GUI/coregui/Views/JobWidgets/ScientificSpinBox.cpp
+++ b/GUI/coregui/Views/JobWidgets/ScientificSpinBox.cpp
@@ -1,0 +1,117 @@
+// ************************************************************************** //
+//
+//  BornAgain: simulate and fit scattering at grazing incidence
+//
+//! @file      GUI/coregui/Views/JobWidgets/ScientificSpinBox.cpp
+//! @brief     Implements class ScientificSpinBox
+//!
+//! @homepage  http://www.bornagainproject.org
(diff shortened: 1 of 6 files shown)


GUI/coregui/Views/JobWidgets/ScientificSpinBox.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+    if (abs_val <= min_val)
+        return false;
+
+    if (abs_val >= upper_switch || abs_val < lower_switch / 2.)
+        return true;
+
+    const double round_val = getRoundedToDecimals(abs_val, decimal_points);
+    return round_val >= upper_switch || round_val < lower_switch;
+}
```
waltervh: Why this double check?

DmitryYurov: Initial idea was to provide unambiguous string representation for values close to upper/lower switch. Looking at the code once again, it's an overkill. Hardly anybody will ever notice.

DmitryYurov:
Will rewrite ScientificSpinBox as a descendant of QAbstractSpinBox. Closing.

PR #703, opened 2018-11-21
Redmine issue 2231: strange behaviour at wavelength distribution

DmitryYurov:


waltervh:
Fantastically clean fix of a bug that managed to hide for a long time...

DmitryYurov:
> Fantastically clean fix of a bug that managed to hide for a long time...
I'm fully and completely flattered. Really.

PR #708, opened 2018-11-23
Exclude zero-probability samples in DistributionTrapezoid (bug #2200)

waltervh:


Core/Parametrization/Distributions.cpp
```diff
@@ -180,17 +183,17 @@ DistributionLorentz::DistributionLorentz(double mean, double hwhm) : m_mean(mean
 double DistributionLorentz::probabilityDensity(double x) const
 {
     if (m_hwhm == 0.0)
```
DmitryYurov: Probably `DoubleEqual(m_hwhm, 0.0)` then?

PR #709, opened 2018-11-26
CsvImportAssistant: Select individual rows to discard + treat multiple spaces as one.

juanmcloaiza:


GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant/CsvImportTable.cpp
```diff
@@ -134,13 +153,32 @@ void CsvImportTable::setData(const csv::DataArray data)
...
+        return;
+    }
+    for (auto row : rows) {
+        bool userDiscard =
+            std::find(m_rowsToDiscard.begin(), m_rowsToDiscard.end(), row) != m_rowsToDiscard.end();
+        if (!userDiscard) {
+            m_rowsToDiscard.push_back(row);
+        } else {
+            m_rowsToDiscard.erase(std::remove(m_rowsToDiscard.begin(), m_rowsToDiscard.end(), row),
+                                  m_rowsToDiscard.end());
```
DmitryYurov: I suppose you have chosen a wrong container type for discarded rows. Update of discarded rows can be done in a clearer way.

GUI/coregui/Views/ImportDataWidgets/CsvImportAssistant/CsvImportTable.cpp
```diff
@@ -234,15 +235,24 @@ void CsvImportTable::runSanityChecks()
 
         // if two consecutive values are non-increasing:
         if (number - greatestNumber < 0) {
-            for (int j = 0; j < nCols; ++j)
-                greyoutCell(i + rowOffset(), j, needsGreyout(i + rowOffset(), j), Qt::red);
+            bool alreadyDiscarded = needsGreyout(i + rowOffset(), jCol);
```
DmitryYurov: You have already checked if the row is discarded. Why do you need to do that once again?

PR #715, opened 2018-11-28
Redmine issue 2141: Fitting material parameters in GUI

DmitryYurov:
The pull-request takes into account previous reviewer remarks.
The issue 2116 still needs to be solved separately, since it requires some changes and wiring
in `FitParameterItem`


gpospelov:
Great, thanks!

PR #723, opened 2018-11-30
Redmine issue 2116: fixes to fit parameter items

DmitryYurov:


GUI/coregui/Models/FitParameterItems.cpp
```diff
@@ -144,6 +144,22 @@ AttLimits FitParameterItem::attLimits() const
...
+    double value = getItemValue(P_START_VALUE).toDouble();
+    double min = getItemValue(P_MIN).toDouble();
+    double max = getItemValue(P_MAX).toDouble();
+
+    if (isLowerLimited())
+        return min <= value;
+    if (isUpperLimited())
+        return value <= max;
+    return min <= value && value <= max;
+}
```
waltervh: What if isLowerLimited() and value > max?

PR #762, opened 2019-02-06
break dll archive into 4 zip files, for gsl, tiff, cblas, fftw3

jwuttke:


jwuttke:
This addresses part of #2163

PR #812, opened 2019-05-10
Treat evey file as ASCII unless it has a specific BA recognized extension.

juanmcloaiza:
This pull request addresses Feature #2311 issue (http://apps.jcns.fz-juelich.de/redmine/issues/2311):

> Situation: [A user] brings *.dat file from experiment which contains simple 2D ascii map.
BornAgain refuses to read it. If we rename file to *.txt it reads it without problems.

> GUI allows for importing data matrix from text files with .txt extension. However, some MLZ instruments save this matrix in the text files with other extensions. For example, KWS3 saves matrix in the text files with extension .det.

> For an average GUI user it is quite some work to rename several tens of files to be able to load them in GUI.

> I think, this constraint should be relaxed. We should not determine the file type by extension.

In this pull request, the logic is changed: Files with extensions other than the BA recognized ones (.int, .tif, .gz, ...) are  treated as ASCII matrices of numbers. Data formats such as .mp3 or .png will still be treated as ASCII and an error will be thrown at the moment of trying to read them.

### DIFF
--- a/Core/InputOutput/OutputDataReadStrategy.cpp
+++ b/Core/InputOutput/OutputDataReadStrategy.cpp
@@ -119,13 +119,20 @@ OutputData<double>* OutputDataReadNumpyTXTStrategy::readOutputData(std::istream&
     std::string line;
     std::vector<std::vector<double>> data;
 
+    //Read numbers from input stream:
     while( std::getline(input_stream, line) ) {
-        if(line.empty() || line[0] == '#')
+        if(line.empty() || !isdigit(line[0]))
             continue;
-        std::vector<double> data_in_row = DataFormatUtils::parse_doubles(line);
-        data.push_back(data_in_row);
+
+        try {
+            std::vector<double> dataInRow = DataFormatUtils::parse_doubles(line);
+            data.push_back(dataInRow);
+        } catch (...) {
+            continue;
+        }
     }
-    // validating
+
+   // validating
     size_t nrows = data.size();
     size_t ncols(0);
     if(nrows) ncols = data[0].size();
--- a/Core/InputOutput/OutputDataReadStrategy.cpp
+++ b/Core/InputOutput/OutputDataReadStrategy.cpp
@@ -22,12 +22,31 @@
 #include <stdexcept> // need overlooked by g++ 5.4
 #include <map>
 
+namespace{
+inline std::string trim(const std::string& str,
+                        const std::string& whitespace = " \t")
+{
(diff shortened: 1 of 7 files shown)


Core/InputOutput/OutputDataReadStrategy.cpp
```diff
@@ -115,13 +115,20 @@ OutputData<double>* OutputDataReadNumpyTXTStrategy::readOutputData(std::istream&
     std::string line;
     std::vector<std::vector<double>> data;
 
+    //Read numbers from input stream:
     while( std::getline(input_stream, line) ) {
-        if(line.empty() || line[0] == '#')
+        if(line.empty() || !isdigit(line[0]))
```
waltervh: Please remove whitespace at the beginning of a line before parsing it

Core/InputOutput/OutputDataWriter.cpp
```diff
@@ -36,9 +36,7 @@ void OutputDataWriter::writeOutputData(const OutputData<double>& data)
         throw Exceptions::NullPointerException("OutputDataWriter::getOutputData() ->"
                                                " Error! No read strategy defined");
     std::ofstream fout;
-    std::ios_base::openmode openmode = std::ios::out;
-    if(DataFormatUtils::isBinaryFile(m_file_name))
-        openmode = std::ios::out | std::ios_base::binary;
+    std::ios_base::openmode openmode = std::ios::out | std::ios_base::binary;
```
waltervh: Please revert to text mode in case of an ascii file. Otherwise newlines might be written wrongly on non POSIX OS's

Core/InputOutput/OutputDataReader.cpp
```diff
@@ -36,9 +36,7 @@ OutputData<double>* OutputDataReader::getOutputData()
             "OutputDataReader::getOutputData() -> Error! No read strategy defined");
 
     std::ifstream fin;
-    std::ios_base::openmode openmode = std::ios::in;
-    if (DataFormatUtils::isBinaryFile(m_file_name))
-        openmode = std::ios::in | std::ios_base::binary;
+    std::ios_base::openmode openmode = std::ios::in | std::ios_base::binary;
```
waltervh: The same applies here (binary flag should only be set when it is a tiff or compressed file)

juanmcloaiza:
The build is failing, I'll create a new PR when I fix it.

juanmcloaiza:
I think is fixed now...

PR #816, opened 2019-05-13
Add function ba.mpl_settings to overwrite MatPlotLib defaults.

jwuttke:


jwuttke:
Problem shall be solved differently.

PR #817, opened 2019-05-13
Treat evey file as ASCII unless it has a specific BA recognized extension.

juanmcloaiza:
This pull request addresses Feature #2311 issue (http://apps.jcns.fz-juelich.de/redmine/issues/2311):

> Situation: [A user] brings *.dat file from experiment which contains simple 2D ascii map.
BornAgain refuses to read it. If we rename file to *.txt it reads it without problems.

> GUI allows for importing data matrix from text files with .txt extension. However, some MLZ instruments save this matrix in the text files with other extensions. For example, KWS3 saves matrix in the text files with extension .det.

> For an average GUI user it is quite some work to rename several tens of files to be able to load them in GUI.

> I think, this constraint should be relaxed. We should not determine the file type by extension.

In this pull request, the logic is changed: Files with extensions other than the BA recognized ones (.int, .tif, .gz, ...) are  treated as ASCII matrices of numbers. Data formats such as .mp3 or .png will still be treated as ASCII and an error will be thrown at the moment of trying to read them.


juanmcloaiza:
Sorry, something is wrong with all those commits being included in this PR... closing it now!

PR #824, opened 2019-05-17
Provide proper behaviour for Simulation::intensityMapSize

DmitryYurov:
Previously it was returning the number of simulation elements for all simulations except for specular one. It is erroneous for off-spec and depth probe. In the case of GISAS it neglected masked areas.

### DIFF
--- a/Core/Simulation/DepthProbeSimulation.cpp
+++ b/Core/Simulation/DepthProbeSimulation.cpp
@@ -102,7 +102,7 @@ const IAxis* DepthProbeSimulation::getZAxis() const
 
 size_t DepthProbeSimulation::intensityMapSize() const
 {
-    if (!m_z_axis || m_alpha_axis)
+    if (!m_z_axis || !m_alpha_axis)
         throw std::runtime_error("Error in DepthProbeSimulation::intensityMapSize: attempt to "
                                  "access non-initialized data.");
     return m_z_axis->size() * m_alpha_axis->size();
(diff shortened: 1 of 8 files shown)


Core/Simulation/DepthProbeSimulation.cpp
```diff
@@ -100,6 +100,14 @@ const IAxis* DepthProbeSimulation::getZAxis() const
...
 }
 
+size_t DepthProbeSimulation::intensityMapSize() const
+{
+    if (!m_z_axis || m_alpha_axis)
+        throw std::runtime_error("Error in DepthProbeSimulation::intensityMapSize: attempt to "
+                                 "access non-initialized data.");
+    return m_z_axis->size() * m_alpha_axis->size();
+}
+
```
waltervh: should read:
if (!m_z_axis || !m_alpha_axis) ...

Core/Simulation/DepthProbeSimulation.cpp
```diff
@@ -100,6 +100,14 @@ const IAxis* DepthProbeSimulation::getZAxis() const
     return m_z_axis.get();
 }
 
+size_t DepthProbeSimulation::intensityMapSize() const
+{
+    if (!m_z_axis || m_alpha_axis)
```
waltervh:
```suggestion
    if (!m_z_axis || !m_alpha_axis)
```

DmitryYurov: Well, it's embarrassing

waltervh: Happens to the best of us :)

PR #827, opened 2019-05-27
Bug 2370 [Reflectometry] - Crash for non-sensible values of theta.

juanmcloaiza:
This Pull Request fixes bug 2370 (http://apps.jcns.fz-juelich.de/redmine/issues/2370):

> To reproduce:
> 
> 1. Instrument view: Choose the Specular Instrument; In the Beam Parameters section --> Inclination Angles [deg] --> Min = 0.0, Max = 91 (i.e. a non-sensible value).
> 2. Sample View: Create any Multilayer sample (I chose the example "Multilayer With Correlated Roughness").
> 3. Simulation View: Hit "Run Simulation".
> 4. Jobs View: The figure shows a plot "alpha_i [deg] vs Signal [a.u.]"; hit "Properties" on top of the plot; on the pane that appears at the right, change axes units to q-space or to radians.
> 5. Don't call it a crash; call it a shortcut for closing the application without being asked whether to save the changes :)
> 
> Possible ways to go (in no particular order):
> 
> A. Impose that the value inserted in (1) is smaller than 90.
> B. Trust that users won't ever set such large values, as they don't make sense anyway.
> C. The error thrown is a std::runtime_error; what(): Error in UnitConverter1D: input axis range is out of bounds . Catch it and give feedback to users through a message box.
> 
> I'd personally go for A.

The fix was done following a logic already present inside BornAgain: Let users write any value inside the Instrument View, but do not run the simulation if the values entered do not make sense. Instead, communicate this information in the small bottom left dock inside the Jobs View.

### DIFF
--- a/Core/Instrument/AngularSpecScan.cpp
+++ b/Core/Instrument/AngularSpecScan.cpp
@@ -22,6 +22,7 @@
 #include "RealLimits.h"
 #include "ScanResolution.h"
 #include "SpecularSimulationElement.h"
+#include "Units.h"
 
 namespace {
 std::vector<std::vector<double>>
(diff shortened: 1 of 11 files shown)


juanmcloaiza:
I was adding changes from another pull request... I'll pick the relevant commits and reopen...

PR #828, opened 2019-05-27
Fix Bug 2370 [Reflectometry] - Crash for non-sensible values of theta.

juanmcloaiza:
This Pull Request fixes bug 2370 (http://apps.jcns.fz-juelich.de/redmine/issues/2370):

> To reproduce:
> 
> 1. Instrument view: Choose the Specular Instrument; In the Beam Parameters section --> Inclination Angles [deg] --> Min = 0.0, Max = 91 (i.e. a non-sensible value).
> 2. Sample View: Create any Multilayer sample (I chose the example "Multilayer With Correlated Roughness").
> 3. Simulation View: Hit "Run Simulation".
> 4. Jobs View: The figure shows a plot "alpha_i [deg] vs Signal [a.u.]"; hit "Properties" on top of the plot; on the pane that appears at the right, change axes units to q-space or to radians.
> 5. Don't call it a crash; call it a shortcut for closing the application without being asked whether to save the changes :)
> 
> Possible ways to go (in no particular order):
> 
> A. Impose that the value inserted in (1) is smaller than 90.
> B. Trust that users won't ever set such large values, as they don't make sense anyway.
> C. The error thrown is a std::runtime_error; what(): Error in UnitConverter1D: input axis range is out of bounds . Catch it and give feedback to users through a message box.
> 
> I'd personally go for A.

The fix was done following a logic already present inside BornAgain: Let users write any value inside the Instrument View, but do not run the simulation if the values entered do not make sense. Instead, communicate this information in the small bottom left dock inside the Jobs View.

### DIFF
--- a/Core/Instrument/AngularSpecScan.cpp
+++ b/Core/Instrument/AngularSpecScan.cpp
@@ -22,6 +22,7 @@
 #include "RealLimits.h"
 #include "ScanResolution.h"
 #include "SpecularSimulationElement.h"
+#include "Units.h"
 
 namespace {
 std::vector<std::vector<double>>
@@ -170,6 +171,17 @@ void AngularSpecScan::setAbsoluteAngularResolution(const RangedDistribution& dis
     std::unique_ptr<ScanResolution> resolution(
         ScanResolution::scanAbsoluteResolution(distr, std_dev));
     setAngleResolution(*resolution);
+}
+
+bool AngularSpecScan::isWithinValidRange() const
+{
+    double min = coordinateAxis()->getMin();
+    double max =  coordinateAxis()->getMax();
+
+    if (min < 0.0 || max >= 90)
+        return false;
+
+    return true;
 }
 
 std::vector<double> AngularSpecScan::footprint(size_t start, size_t n_elements) const
--- a/Core/Instrument/AngularSpecScan.cpp
+++ b/Core/Instrument/AngularSpecScan.cpp
@@ -178,7 +178,7 @@ bool AngularSpecScan::isWithinValidRange() const
     double min = coordinateAxis()->getMin();
     double max =  coordinateAxis()->getMax();
 
-    if (min < 0.0 || max >= 90)
+    if (min < 0.0 || Units::rad2deg(max) >= 90)
         return false;
 
     return true;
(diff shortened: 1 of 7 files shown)


Tests/UnitTests/Core/Fresnel/SpecularSimulationTest.cpp
```diff
@@ -89,15 +89,13 @@ TEST_F(SpecularSimulationTest, CloneOfEmpty)
 TEST_F(SpecularSimulationTest, SetAngularScan)
 {
     SpecularSimulation sim;
-
-    AngularSpecScan scan(1.0, std::vector<double>{1.0, 3.0});
+    AngularSpecScan scan(1.0, std::vector<double>{1.0 * Units::deg, 3.0 * Units::deg});
```
gpospelov: What is the purpose of this change? The test behavior didn't change.

Core/Instrument/AngularSpecScan.cpp
```diff
@@ -172,6 +173,17 @@ void AngularSpecScan::setAbsoluteAngularResolution(const RangedDistribution& dis
     setAngleResolution(*resolution);
 }
 
+bool AngularSpecScan::isWithinValidRange() const
```
gpospelov: To my opinion, the method (and hardcoded check in it) is unnecessary. If you think, that parameters are wrong and nonphysical, it is better to throw in constructor.

juanmcloaiza: True, the motivation behind this change was to mimic what was already in place before, to avoid creating a new logic. Now that I know, I'll take it into account and improve this PR.

Tests/UnitTests/Core/Fresnel/SpecularSimulationTest.cpp
juanmcloaiza: Purpose A: Correctness.

3 radians are almost 180 deg. When the machinery that checks for the validity of the ranges is in place, the test will fail, as 3 radians (180 deg,) are certainly outside the range of validity of the incident angular range.

Purpose B: Consistency and aesthetics.

The rest of the tests in this file are written in degrees, not in radians.

juanmcloaiza:
I'll improve the solution as requested, dropping some commits in favor of new ones.

PR #831, opened 2019-05-28
nomenclature: "rectangular" -> "flat detector" ; GUI: rename "Import" button to "Data"

juanmcloaiza:
This pull request addresses the following redmine issues:

 - Feature #2355 - nomenclature: "rectangular" -> "flat detector" (http://apps.jcns.fz-juelich.de/redmine/issues/2355): 

> For the big renaming release: the correct opposition is "spherical" vs "flat". Rectangular, in there respective coordinates, they are both.

 - Feature #2365 - GUI: rename "Import" button to "Data" (http://apps.jcns.fz-juelich.de/redmine/issues/2365):

> Also change the tooltip.
This button gives access to the "Import 2D" and "Import 1D"
buttons, but also to the imported data display. Therefore,
"Data" is the more comprehensive name.


waltervh:
This kind of huge renaming, which impacts almost all user scripts, will need to be postponed to a later release. We could bundle all major API breaking renaming (see also #1569) in a single release with higher major version number (e.g. BornAgain 2.0).

PR #834, opened 2019-06-05
Bug Fix: GUI crashes with masks and 90deg beam inclination

juanmcloaiza:
This pull request fixes the Bug #2347 (http://apps.jcns.fz-juelich.de/redmine/issues/2347):

> Reproduce:
Load bug.pro in attachment
In import view: change units to qspace and back
In instrument view: set background to none

Beware: This pull request uses two tricks that are a bit dirty, namely, 

1. On 0/0 division, return zero, 
2. On cos(+/- 90deg) return cos(+/- (90 - 1e-14) deg)

The more correct approach would be to prevent users to use 90 deg. as beam inclination angle, as the Cartesian-Polar coordinate transformation becomes singular at the poles. The only reason not to take this approach is simply because writing 90deg looks prettier than writing 89.9999. In the end, it depends on how much precision is actually required.

### DIFF
--- a/Core/Instrument/IntensityDataFunctions.cpp
+++ b/Core/Instrument/IntensityDataFunctions.cpp
@@ -26,6 +26,7 @@
 #include "SimulationResult.h"
 #include "UnitConverterUtils.h"
 #include <math.h>
+#include <limits.h>
 
 //! Returns sum of relative differences between each pair of elements:
 //! (a, b) -> 2*abs(a - b)/(|a| + |b|)      ( and zero if  a=b=0 within epsilon )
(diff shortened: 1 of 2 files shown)


waltervh:
I'm afraid this is a bit of a hack. The real problem is situated in the messy way we convert units in MaskUnitsConverter.cpp. Which means that the mess there should be cleaned up so the fix would become obvious (which it clearly isn't now...)

juanmcloaiza:
> I'm afraid this is a bit of a hack.

It is! So I'll close this PR and see if with the feedback you just gave me I can fix this properly.

PR #837, opened 2019-06-14
Switch GUI to the new objective metrics

DmitryYurov:
Python API extensions will be handled in a separate PR. All the reviewer's remarks were taken into account.


gpospelov:
Exemplary, thank you!

PR #841, opened 2019-07-02
Core refactoring: preprocess MultiLayer into new object and reduce its interface

waltervh:


gpospelov:
No objections from my side, only admiration. Giving place to the next reviewer...

PR #846, opened 2019-07-09
New button in ImportDataView selector

gpospelov:
Attempt to solve the problem with obscured rotate-dataset-action by adding hamburger-like button.


GUI/coregui/Views/ImportDataWidgets/RealDataSelectorHBar.h
```diff
@@ -0,0 +1,39 @@
...
+
+//! Toolbar on top of selector tree with hamburger-like menu button.
+
+class BA_CORE_API_ RealDataSelectorHBar : public QToolBar
+{
+    Q_OBJECT
+public:
+    RealDataSelectorHBar(RealDataSelectorActions* actions, QWidget* parent = nullptr);
+
+private slots:
```
DmitryYurov: I would assume that specifying this function as a slot explicitly is not required after transition to C++11-connections.

PR #852, opened 2019-07-24
Fix application crash on attempt to write project into non-ascii directory under Windows

gpospelov:


waltervh:
Very nicely done!

PR #853, opened 2019-07-26
Simplify CMake search of fftw3

gpospelov:
Also make CMake rising an error when only static version of fftw3 is present on the system.

### DIFF
--- a/cmake/generic/modules/FindFFTW.cmake
+++ b/cmake/generic/modules/FindFFTW.cmake
@@ -1,73 +1,46 @@
 # Find the FFTW includes and library.
-# 
+#
 # This module defines
 # FFTW_INCLUDE_DIR, where to locate fftw3.h file
 # FFTW_LIBRARIES, the libraries to link against to use fftw3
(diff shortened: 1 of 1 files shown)


waltervh:
Nice cleaning up!

cmake/generic/modules/FindFFTW.cmake
```diff
@@ -1,73 +1,52 @@
...
-        NO_DEFAULT_PATH
-    )
+if (NOT APPLE AND NOT WIN32)
+    set(CMAKE_FIND_LIBRARY_SUFFIXES .so) # require shared version of library
+endif()
 
+find_library(FFTW_LIBRARY NAMES ${FFTW_NAMES} QUIET)
+set(CMAKE_FIND_LIBRARY_SUFFIXES ${ba_ORIG_CMAKE_FIND_LIBRARY_SUFFIXES})
+
+if (WIN31)
```
DmitryYurov: Definitely, WIN32 was meant

PR #857, opened 2019-07-29
Add header blocks, update @file entries, add/correct/extend Doxygen comments, simplify hash function classes

jwuttke:


Core/Multilayer/Hash2Doubles.cpp
```diff
@@ -13,6 +13,7 @@
 // ************************************************************************** //
 
 #include "Hash2Doubles.h"
+#include <functional> // hash
```
DmitryYurov: I would probably remove trivial comments like `// hash`.

jwuttke:
Uncontroversial part moved to #863

PR #860, opened 2019-07-30
Switch to the new divergence machinery in GUI/specular and remove old tests

DmitryYurov:


GUI/coregui/Models/TransformToDomain.cpp
```diff
@@ -259,4 +260,25 @@ void setParameterDistributionToSimulation(const std::string& parameter_name,
     if (P_par_distr)
         simulation.addParameterDistribution(*P_par_distr);
 }
+
+std::unique_ptr<ScanResolution> createScanResolution(const SessionItem* item)
+{
+    auto beam_item = dynamic_cast<const BeamDistributionItem*>(item);
+    if (!beam_item)
+        return nullptr;
```
gpospelov: I would personally prefer return {}; in such cases.

cmake/generic/modules/FindFFTW.cmake
```diff
@@ -24,7 +24,7 @@ endif()
 find_library(FFTW_LIBRARY NAMES ${FFTW_NAMES} QUIET)
 set(CMAKE_FIND_LIBRARY_SUFFIXES ${ba_ORIG_CMAKE_FIND_LIBRARY_SUFFIXES})
 
-if (WIN31)
+if (WIN32)
```
gpospelov: This was nice catch!

PR #862, opened 2019-07-30
Fix memory leakage both in GUI and PythonAPI

gpospelov:
Leak summary
`
==79988==    definitely lost: 38,162 bytes in 988 blocks
==79988==    indirectly lost: 42,095,920 bytes in 3,206 blocks
`
After
`
==14538==    definitely lost: 20,978 bytes in 454 blocks
==14538==    indirectly lost: 2,184 bytes in 15 blocks
`


DmitryYurov:
Thanks for fixing. Shame on me.

PR #863, opened 2019-07-30
Add header blocks, update @file entries, add/correct/extend Doxygen comments

jwuttke:
Split off from PR #857.


gpospelov:
I would personally remove both, "file" and "brief" entries from the header completely. The field "brief" doesn't provide much additional information with respect to class description right after. The field "file" contains useless path and requires support. 
To my opinion, current directory structure can be better and should be reorganized in the future in accordance with the CMake's idea of public/private API. So removing path now will save us from useless and massive files update later on.

jwuttke:
@gpospelov: can be discussed this in a group meeting. As long as we have these fields, they should be correct.

PR #864, opened 2019-07-30
Make hash classes local, and rm member variables

jwuttke:
To be considered later:
  * Depends on PR #863.
  * Await forthcoming release.


gpospelov:
Don't worry about failed build. I was cleaning up a long queue on Appveyor. Will re-launch after previous pull request merged.

PR #878, opened 2019-09-18
Fix O(n_layer^2) bug

DmitryYurov:
Before:

n_layers = 10,	 time = 4 ms
n_layers = 100,	 time = 64 ms
n_layers = 200,	 time = 349 ms
n_layers = 500,	 time = 4547 ms
n_layers = 1000,	 time = 34026 ms

After:

n_layers = 10,	 time = 4 ms
n_layers = 100,	 time = 28 ms
n_layers = 200,	 time = 52 ms
n_layers = 500,	 time = 210 ms
n_layers = 1000,	 time = 646 ms

### DIFF
--- a/Core/Multilayer/MultiLayer.cpp
+++ b/Core/Multilayer/MultiLayer.cpp
@@ -105,11 +105,13 @@ void MultiLayer::setExternalField(kvector_t ext_field)
 std::vector<const INode*> MultiLayer::getChildren() const
 {
     std::vector<const INode*> result;
-    for (auto p_layer : m_layers) {
-        result.push_back(p_layer);
-        auto i_layer = MultiLayerUtils::IndexOfLayer(*this, p_layer);
-        if (const LayerInterface* p_interface =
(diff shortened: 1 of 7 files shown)


gpospelov:
Very cool! Such nasty and long playing bug is finally resolved. Thank you!

DmitryYurov:
Thank you, sir! It's just our job to make the world a better place.

juanmcloaiza:
Great Dmitry!! 👌🏾🎉

juanmcloaiza:
Good news: Timings are indeed much better than before.

Bad news: Timings are still scaling as $O(N^2)$, though it is only noticeable for large numbers of layers.

From the numbers @DmitryYurov provided, the trend is not clearly noticeable,

image

but carrying on the analysis with more and more layers, the trend is clearer:

image

DmitryYurov:
Yes, indeed something is still slowing down the computation on samples with an extremely big number of layers. Looking into it right now.

PR #891, opened 2020-01-21
Revert "UserManual/Scattering.tex: Correct eq. 1.22, add missing r'^2"

waltervh:
Reverts scgmlz/BornAgain#890

Sorry for the confusion, but this is not an equation. The first two 'equation' signs denote approximate equality. In this case, the square of r' can be neglected, which is exactly what you want to achieve in the far field approximation.


gpospelov:
Thanks for being around and keeping an eye on us

waltervh:
You're welcome ;-)

PR #908, opened 2020-03-27
Make strategy for specular computation switchable

rbeerwerth:
Prepare the codebase to add Nevot-Croce Roughness model

### DIFF
--- a/auto/Wrap/libBornAgainFit_wrap.cpp
+++ b/auto/Wrap/libBornAgainFit_wrap.cpp
@@ -5800,7 +5800,7 @@ SWIG_AsVal_std_complex_Sl_double_Sg_  (PyObject *o, std::complex<double>* val)
 
 
 SWIGINTERNINLINE PyObject*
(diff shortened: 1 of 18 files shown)


auto/Wrap/libBornAgainFit_wrap.cpp
```diff
@@ -5800,7 +5800,7 @@ SWIG_AsVal_std_complex_Sl_double_Sg_  (PyObject *o, std::complex<double>* val)
 
 
 SWIGINTERNINLINE PyObject*
-SWIG_From_std_complex_Sl_double_Sg_  (/*@SWIG:/usr/local/share/swig/3.0.12/typemaps/swigmacros.swg,104,%ifcplusplus@*/
+SWIG_From_std_complex_Sl_double_Sg_  (/*@SWIG:/usr/share/swig3.0/typemaps/swigmacros.swg,104,%ifcplusplus@*/
```
gpospelov: I forgot to mention. To avoid different SWIG's messages from different developers we are using same version of swig installed into same location. Could you please install (at some point) SWIG version 3.0.12 into /usr/local, from source code. Then this difference will not appear anymore.

rbeerwerth: will do, thx!

PR #909, opened 2020-03-30
Add Nevot-Croce roughness model

rbeerwerth:


gpospelov:
Looks good, thank you.

PR #917, opened 2020-04-02
Add testcase for Nevot-Croce roughness + mod. RTTest

rbeerwerth:


Core/StandardSamples/MultiLayerWithRoughnessBuilder.h
```diff
@@ -26,7 +26,7 @@ class BA_CORE_API_ MultiLayerWithRoughnessBuilder : public IMultiLayerBuilder
 {
 public:
     MultiLayerWithRoughnessBuilder();
-    MultiLayer* buildSample() const;
+    virtual MultiLayer* buildSample() const;
```
gpospelov, line 29: We are not consistent in BornAgain, but I would suggest to drop the keyword "virtual" in all derived classes and use "override".

Core/Multilayer/SpecularScalarStrategy.cpp
```diff
@@ -95,21 +95,17 @@ bool SpecularScalarStrategy::calculateUpFromLayer(std::vector<ScalarRTCoefficien
...
-        // this avoids an overflow in the backwards propagation
-        // the used factors are stored in order to correct the amplitudes
-        // in forward direction at the end of the computation
-
-        if (std::isinf(std::norm(coeff[i].t_r(0)))){
-//            throw std::runtime_error("Unexpected inf");
-            // catching this inf is not correctly covered by existing tests
-            // it is possible to pass all tests by moving the normalization before
-            // this check and hence corrupting all amplitudes
+        if (std::isinf(std::norm(coeff[i].t_r(0))) || std::isnan(std::norm(coeff[i].t_r(0)))){
```
gpospelov, line 98: Nice catch

PR #937, opened 2020-05-06
Modern way of finding Python on a system

gpospelov:
+ Provide a consistent way of finding Python interpreter, Python libs, and NumPy on a system with multiple Python installations.
+ New docker container to use for the build (Ubuntu-focal April 2020: newer CMake, newer Qt).
+ CMake minimum required version now 3.14

Two strange issues fixed:
+  The behavior of Qt5 + Docker  + Appveyr while finding *.qrc files on disk.
+ Strange way of parsing Python_EXECUTABLE variables on Appveyor with the mess in "/" against "\\".


rbeerwerth:
sweet, working nicely!

cmake/bornagain/modules/BornAgainConfiguration.cmake
```diff
@@ -162,10 +162,6 @@ configure_file(${TEMPLATE_DIR}/thisbornagain.csh.in
 
 if(BUILD_DEBIAN)
     set(CMAKE_INSTALL_PREFIX "/usr")
-    execute_process(COMMAND "${PYTHON_EXECUTABLE}" -c
```
juanmcloaiza, line 165: Why isn't this needed anymore?

juanmcloaiza, line 165: This line was intended to define the variable `PYTHON_SITE_PACKAGES`, which is now dropped and instead, int the rest of the files was replaced by `Python_SITELIB`. Where did you define this new variable? What's the logic behind the new way of defining it?

gpospelov: This is defined [here](https://cmake.org/cmake/help/git-stage/module/FindPython.html).

juanmcloaiza: to the point, thank you!

PR #939, opened 2020-05-08
Core: FormFactorFullSpheroid: analytic instead of numeric computation

jwuttke:
Following a suggestion by Matt Thompson, simplified the computation and updated the documentation.

Also some copy editing on FFCatalog; improvements to the index.

Replotted all simulation outcomes on reduced intensity scale (8 instead of 10 decades) and with color map 'inferno' as in paper.


jwuttke:
removed 'WIP' tag: ready for merge

PR #948, opened 2020-05-14
Change FormFactorDot. Suppress warnings. Edit FFCatalog.

jwuttke:
- Suppress warnings from swig-generated code
- Eliminate several other warnings
- FFCatalog: Provide short intro to ripples; start chapter on soft particles
- FormFactorDot: Add 'radius' argument to get dimensions and intensity scaling right


Tests/Functional/GUI/GUIStandardTest/GUIStandardTest.cpp
```diff
@@ -49,14 +49,11 @@ std::unique_ptr<Simulation> createDomainSimulation(const Simulation& origin)
 bool GUIStandardTest::runTest()
 {
     m_reference_simulation->runSimulation();
-    auto ref_result = m_reference_simulation->result();
+    const SimulationResult& ref_result = m_reference_simulation->result();
```
gpospelov, line 52: I find the syntax with `const` reference to a temporary object more confusing. It is allowed, but confusing, to my taste. My suggestion would be to stick to [almost always auto](https://blog.galowicz.de/2016/03/23/const_reference_to_temporary_object/) in newly written code, while leaving old code intac.

Core/Tools/PythonCore.h
```diff
@@ -28,7 +28,11 @@
 #define NO_IMPORT_ARRAY
 #include <numpy/arrayobject.h>
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
gpospelov: That's nice that old Macros.h is removed, it was generating warnings in my case too. For the future, however, we could consider another, more modern version of similar macros, to centralize Warning suppression in the compiler independent way. There is a nice [post](https://www.fluentcpp.com/2019/08/30/how-to-disable-a-warning-in-cpp/) about the problem.

Core/CMakeLists.txt
```diff
@@ -66,6 +66,14 @@ if(BORNAGAIN_PYTHON)
             list(APPEND SWIG_FLAGS "-I${dir}")
         endforeach(dir)
 
+        if ((CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+                OR (CMAKE_CXX_COMPILER_ID STREQUAL "GNU"))
+            # suppress warnings from auto-generated code (last updated for Swig 4.0.1)
+            set_source_files_properties(${AUTO_DIR}/libBornAgainCore_wrap.cpp
+                PROPERTIES COMPILE_OPTIONS
+                "-Wno-unused-parameter;-Wno-missing-field-initializers;-Wno-sometimes-uninitialized;-Wno-deprecated-declarations")
+        endif()
```
gpospelov, line 75: This is nice!

GUI/coregui/Views/SampleDesigner/DesignerHelper.cpp
```diff
@@ -37,7 +37,7 @@ QGradient DesignerHelper::getLayerGradient(const QColor& color, const QRectF& re
     result.setColorAt(0.5, c.light(200));
     result.setColorAt(1, c.dark(150));
 #endif
-    return result;
+    return std::move(result);
```
gpospelov, line 40: Changes do not match the commit message. Also move is not necessary in this context, to my opinion.

Tests/Functional/GUI/GUIStandardTest/GUIStandardTest.cpp
gpospelov: Gave a wrong reference to [almost always auto](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/)

GUI/coregui/Views/SampleDesigner/DesignerHelper.cpp
jwuttke: Was prompted by clang warning.

gpospelov: Good to know, will investigate, but probably, clang is wrong :). Thanks.

PR #952, opened 2020-05-19
Simplify integrator

jwuttke:
New integration functions take the integrand as a lambda function. No more need for templates. Type-safe and easy to read. Measured: exactly as fast as the superseded variant.


jwuttke:
closing while fixing several minor problems

jwuttke:
Minor problems were either inexistent, or unconnected with this PR.

PR #960, opened 2020-05-29
Disable Manhattan Style

gpospelov:
+ Get rid of custom Qt style borrowed from Qt-creator
+ That fixes many visual glitches for certain x-managers.
+ Replace custom made icons with [Material Design](https://materialdesignicons.com/) set.
+ Enable high DPI support. That improves icons rendering under Linux for high DPI screens. Mac and Windows have to be still validated.


rbeerwerth:
Looks pretty good on Linux!
Somehow on Windows there seem to be some scaling issues.
screen1
The labels and contents of the text boxes do not seem to scale, while the menu bar is fine. The text next to the icons does not scale either, while the left bar does. So does the list of particles and the tree and property view.
screen2
The button in the recent files dialog doesn't scale on both Linux + Windows.

PR #972, opened 2020-06-15
FormFactor: add CantellatedCube

jwuttke:
and in FFCatalog outcomment SoftParticles section, which is not yet ready for release.


rbeerwerth:
Is it intended that the new form factor does not appear in the GUI?
Or am i missing something?

jwuttke:
> Is it intended that the new form factor does not appear in the GUI?

Yes, I am aware that the new form factor is still missing in the GUI. This should not prevent us from making it available to Python users.

Also, bringing a new form factor to the GUI is painful. Let's wait for the new GUI and then see how things can be streamlined.

And finally, this reminds me that we also need artwork, http://apps.jcns.fz-juelich.de/redmine/issues/2464.

rbeerwerth:
okay, thx!

PR #982, opened 2020-06-18
CHANGELOG: start new section

jwuttke:


gpospelov:
I have migrated to another Eigen repository http://apps.jcns.fz-juelich.de/redmine/issues/2447. 
I think you have to re-init your submodule. Like now you have changed somehow commit of Eigen pointing back to the old place while submodule itself is configured for new URL. I have to check, how to properly re-init the submodule in such a case. Your PR shouldn't have changes in Eigen dir.

PR #984, opened 2020-06-22
CMake: modernize in view of vcpkg

jwuttke:
- rm FindGSL.cmake, supplied by CMake 3.14
- rename FindTIFF.cmake to FindTIFFxx to differentiate from the CMake supplied module
- rename FindFFTW.cmake to FindFFTW3
- replace *_LIBRARY by *_LIBRARIES
to prepare for VCPKG.
This resolves http://apps.jcns.fz-juelich.de/redmine/issues/2465


gpospelov:
+ Windows installer seems to be broken

fftw-problem

+ Linux CMake is working but issue warnings
```
-- Looking for libtiff (use -DBORNAGAIN_TIFF_SUPPORT=OFF to disable)
CMake Warning (dev) at /usr/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:272 (message):
  The package name passed to `find_package_handle_standard_args` (TIFF) does
  not match the name of the calling package (TIFFxx).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  cmake/generic/modules/FindTIFFxx.cmake:48 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  cmake/bornagain/modules/SearchInstalledSoftware.cmake:46 (find_package)
  CMakeLists.txt:53 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found TIFF: /usr/lib/libtiff.so (found version "4.1.0") 
```

+ Mac OS I will be able to validate only in the evening

jwuttke:
Linux warning fixed.

gpospelov:
Shall we close this pull request? Seems the issue with broken installer needs some time to resolve?

jwuttke:
On which Windows machine did the installer fail to find libfftw3? How are external dependences maintained on that machine?

gpospelov:
On my developer's Windows virtual machine. When I'm checking the installer, I remove all external dependencies from PATH, pretending it is clean Windows. After BornAgain installed, I see that C:\BornAgain-1.17.0\bin directory doesn't contain libtiff and libfftw3, which is normally the case. The installer is also smaller than, for example, on previous pull request - 22.8 Mb .vs. 23.6 in zip. And I run installed BornAgain, it complains about missed fftw.

jwuttke:
So the X.dll (X=libtiff, lifftw3) are on the machine (Where do they come from?), they used to be copied by 'make install' to C:\BornAgain-1.17.0\bin, and this is no longer the case?
That means, our FindX.cmake had some extra functionality that is not normally provided by CMake's own Find modules?

gpospelov:
What can be, is that installer attached to this pull request somehow was not updated and contains results of one of previous build (where everything was broken) and was not updated. I will try again with the recent.

gpospelov:
I think it is Core/CmakeLists.txt 
```
    install(FILES
        ${FFTW_LIBRARY_DLL}
        ${TIFF_LIBRARY_DLL}
        DESTINATION ${destination_lib} COMPONENT Libraries)
```
Where names of DLL are deduced from names of *.lib

I was writing about this here https://jugit.fz-juelich.de/mlz/intern/ba-intern/-/blob/master/release/local-win-vcpkg.md
```
Archive contains *.dll files in bin directory, *.lib files in /lib directory. Our CMake (fftw, tiff parts) expects them in the same directory.
Our FindFFTW and FindTiff (and following code) relies on non-standard variables FFTW_LIBRARY, while toolchain generates standard FFTW3_LIBRARIES variables.
```

jwuttke:
Something unrelated: My patch to FindTIFF.cmake has been accepted by upstream CMake. Therefore I revert from FindTIFFxx t FindTIFF, with 'COMPONENTS CXX' option.

jwuttke:
Back to the Windows problem.
Where in our CMake files is installation to `C:\BornAgain-1.17.0\bin` configured?
What about this proposal: https://stackoverflow.com/a/56514534/1017348? Something like
```
add_library(foo SHARED foo.cpp)
set_target_properties(foo
    PROPERTIES
        LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<CONFIG>/lib
        RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<CONFIG>/bin
)
```

gpospelov:
> Where in our CMake files is installation to C:\BornAgain-1.17.0\bin configured?

In Core/CMakeLists.txt.

```
    install(FILES
        ${FFTW_LIBRARY_DLL}
        ${TIFF_LIBRARY_DLL}
        DESTINATION ${destination_lib} COMPONENT Libraries)
```

> What about this proposal: https://stackoverflow.com/a/56514534/1017348? 

This recipe, as I can see, discusses how to make your own pair of lib/dll and where to put it to conform with the standard location. 
Seems reasonable location definition.

It can be used in the case of external library FFTW3 too I guess. But we probably still have to use `install(FILES)` directive.

However, there is an additional complication in this case - where to find FFWT3 lib/dll before installing it. In current BornAgain we are guessing the location of `dll`  using FFTW(3)_LIBRARIES variable and basing on the known structure of our own archive (where fftw3.lib and fftw3.dll are located in the same directory). This is all hand-made and non-standard.

There are two options, in my opinion.

A) We keep using our old archive but have to fix `dll` copying in the install directive by checking that FFTW3_LIBRARIES are pointing to right `libs`, and our custom FFTW_LIBRARY_DLL are pointing to right `dll` locations.

B) We also change the way of distributing dependencies from custom made archive to VCPKG made archive.
It has it's own standard structure, particularly it has `.dll` in `/bin`, and `.lib` in `/lib`. So basically the way we are finding FFTW_LIBRARY_DLL will be different. Also, I do believe, that this archive doesn't use FindFFTW3.cmake but modern FFTW3Config.cmake. FFTW3_* variables should have the same definition.

jwuttke:
I am pretty confident that the Windows problem is solved. The CMake instructions to install the FFTW3 and TIFF DLLs were actually present in Core/CMakeLists.txt, but had incorrect CMake foreach/list syntax. Now, according to ttps://github.com/scgmlz/BornAgain/pull/984/checks?check_run_id=824128394, "Build package", lines 715 and 716, the DLLs are packed into the installer.

What about MacOS?

gpospelov:
Sorry, I've missed the question. @AlexanderSchober, would you be so kind as to check an installer from [here](https://github.com/scgmlz/BornAgain/actions/runs/153263030) on your macOS. It will be enough just to install/run it and see that the application actually starts.

jwuttke:
Rebased on latest develop, and confirmed on a Windows machine that the installer works, that libfftw3-3 and libtiff DLLs are installed, and that the BornAgain executable starts.

AlexanderSchober:
Tested and approved under MacOsX Catalina

PR #985, opened 2020-06-25
Doxygen: now all variants fully under CMake control

jwuttke:
Three flavors of Doxygen documentation (user, dirgraphs, long) are now configured in uniform way, through command `cmake -DBORNAGAIN_CONFIGURE_DOXY=ON ..`. To build the documentation, run commands `make doxyuser` etc. Or `make doxy` for all flavors.


auto/Doxygen/Doxyfile.long
```diff
@@ -820,9 +820,9 @@ WARN_LOGFILE           =
 # spaces. See also FILE_PATTERNS and EXTENSION_MAPPING
 # Note: If this tag is empty the current directory is searched.
 
-INPUT                  = ./ \
-                         ../../Core \
-                         ../../Fit
+INPUT                  = /G/sw/ba/Doc/Doxygen \
+                         /G/sw/ba/Core \
+                         /G/sw/ba/Fit
```
gpospelov, line 825: What is this path?

auto/Doxygen/Doxyfile.long
```diff
@@ -1134,7 +1134,7 @@ GENERATE_HTML          = YES
 # The default directory is: html.
 # This tag requires that the tag GENERATE_HTML is set to YES.
 
-HTML_OUTPUT            = html/long
+HTML_OUTPUT            = /G/sw/ba/html/long
```
gpospelov, line 1137: Seems hardcoded path

auto/Doxygen/Doxyfile.long
jwuttke: Comes from configuring under my local environment. Many thanks for spotting this stupid mistake.

jwuttke: Auto-generated Doxyfiles now go to CMAKE_BINARY_DIR, and are no longer under version control.

auto/Doxygen/Doxyfile.long
jwuttke: see above

auto/Wrap/libBornAgainFit_wrap.cpp
```diff
@@ -5491,7 +5491,7 @@ SWIG_AsVal_std_complex_Sl_double_Sg_  (PyObject *o, std::complex<double>* val)
 
 
 SWIGINTERNINLINE PyObject*
-SWIG_From_std_complex_Sl_double_Sg_  (/*@SWIG:/usr/local/share/swig/4.0.1/typemaps/swigmacros.swg,104,%ifcplusplus@*/
+SWIG_From_std_complex_Sl_double_Sg_  (/*@SWIG:/usr/share/swig4.0/typemaps/swigmacros.swg,104,%ifcplusplus@*/
```
gpospelov, line 5494: I would suggest sticking to the old agreement, and to use the same version of SWIG (4.0.1 currently) installed in the same directory (/use/local). By that, we do not have `libBornAgainFit_wrap.cpp` changing all the time depending on the way SWIG installed on developer's local machine.

PR #992, opened 2020-07-01
Unit tests: Remove unnecessary destructors

jwuttke:
Remove explicit code for destructors, and some constructors, that are automatically defined by modern C++.
Also unpointerize a few objects.

This saves us 400 LOC.

### DIFF
--- a/Tests/UnitTests/Core/DataStructure/OutputDataIteratorTest.cpp
+++ b/Tests/UnitTests/Core/DataStructure/OutputDataIteratorTest.cpp
@@ -6,56 +6,49 @@ class OutputDataIteratorTest : public ::testing::Test
 {
 protected:
     OutputDataIteratorTest();
-    virtual ~OutputDataIteratorTest();
 
-    OutputData<double>* p_data;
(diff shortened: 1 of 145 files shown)


Tests/UnitTests/Core/DataStructure/OutputDataIteratorTest.cpp
```diff
@@ -6,56 +6,49 @@ class OutputDataIteratorTest : public ::testing::Test
 {
 protected:
     OutputDataIteratorTest();
-    virtual ~OutputDataIteratorTest();
 
-    OutputData<double>* p_data;
+    OutputData<double> _data;
```
rbeerwerth, line 10: i'm confused about the naming convention. p_ doesn't follow what's written down. How about _? Maybe we should talk about this at some point and update internal coding rules

jwuttke: Good point. Please take notes of such observations across the code, and make suggestions for unification.

PR #993, opened 2020-07-02
Added Doxygen class comments. Renamed WallclockTimer <- TimeInterval

jwuttke:
in partial fulfillment of http://apps.jcns.fz-juelich.de/redmine/issues/2392


Fit/Tools/WallclockTimer.h
```diff
@@ -18,22 +18,24 @@
...
     void start();
     void stop();
 
-    //! returns run time in sec.msec
+    //! returns run time in sec.
     double runTime() const;
 
 private:
-    std::unique_ptr<TimeIntervalImp> m_imp;
+    std::unique_ptr<WallclockTimerState> m_state;
```
rbeerwerth, line 38: I believe the ending Imp was introduced to highlight that this implementation follows the PImpl idiom. I wouldn't resist renaming these, but to some extend it hides the idea behind the code, as I think the imp or a impl ending is a very common naming convention, and is also extensively used in BA.

jwuttke: https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes: prefer struct over class if it holds only data, no methods. So in general, I agree with you (prefer "impl" over "imp" though), but here, it's less than an implementation class, it's an internal state only.

PR #995, opened 2020-07-02
WIP: Fit tests now in Fit/test; all tests now under CTest control; no more forced tests during compilation

jwuttke:
Fit tests have been moved from Tests/...Fit... to Fit/test/... - This is a first step towards taking the Multifit library out of BornAgain.

All Unit tests are now under CTest control. See https://blog.kitware.com/dynamic-google-test-discovery-in-cmake-3-10/ for background. There are no more forced tests during compilation. All unit tests have "Unit:" in the name. So they can be run by the command "ctest -R Unit:".


jwuttke:
A number of files from Tests/Functional/TestMachinery had to be duplicated; the copies are in Fit/test/engine. This is not ideal, but can be repaired in the future in two ways:

- Make of those sources a little test support library of its own.
- Get rid of this machinery altogether because it mostly duplicates functionality that is also provided by CTest

jwuttke:
>     * Get rid of this machinery altogether because it mostly duplicates functionality that is also provided by CTest

After inspecting the code I'm quite certain this is the way to go: 
http://apps.jcns.fz-juelich.de/redmine/issues/2474

jwuttke:
Will first commit the easy things in a separate PR. Then simplify the functional test machinery. Only then move Minimizer tests to Fit.

PR #1016, opened 2020-07-29
Generate debian package and upload it as an artifact through github actions.

juanmcloaiza:
These few modifications seem to do the trick; The questionable modification could be that I'm coming back to installing python through apt-get, as opposed to doing it through github actions. This modification was done because github actions install python in a strange directory (`/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/`), which the deb pack-machinery (or my way of using it) failed to find (usually in `/usr/lib/python3/dist-packages/`). However, bornagain compiles and installs successfully without impacting the github actions workflow time.


jwuttke:
Thank you, @juanmcloaiza. I rebased on latest develop. Can be merged as soon as checks have completed.

jwuttke:
Sorry, don't merge yet. We need to manually test whether the resulting Debian package installs properly.

jwuttke:
Resulting Debian package depends on libboost-all-dev. A binary package should depend on library binaries only, not on "-dev" packages.

jwuttke:
Different Debian or Ubuntu versions provide different libgsl versions. Dependence on libgsl23 should be relaxed into `libgsl2 or libgsl23 or libgsl25`.

juanmcloaiza:
By not using the "-dev" dependencies, the debian package is not installable, throwing the following error:

```
The following packages have unmet dependencies:
 bornagain : Depends: libgsl (>= 1.15) but it is not installable
             Depends: libboost-filesystem (>= 1.60.0) but it is not installable
             Depends: libboost-program-options (>= 1.60.0) but it is not installable
             Depends: libboost-regex (>= 1.60.0) but it is not installable
             Depends: libboost-iostreams (>= 1.60.0) but it is not installable
```

Further investigation should be carried out.

jwuttke:
`libgsl` is not a valid package name. Rather use `libgsl2 or libgsl23 or libgsl25`.

juanmcloaiza:
Thank you Joachim, yes, the same problem is found in the rest of the packages I reported in the message that closed the pull request.

PR #1031, opened 2020-08-08
Rename CMake options 'GENERATE..' -> 'CONFIGURE..'

jwuttke:
and omit 'BORNAGAIN_' so that I can use certain shell aliases across projects.


gpospelov:
My personal preference would be to keep `BORNAGAIN` prefix for all our keys. I found it convenient to distinguish keys of one project from the keys of another project while working in a multimodule environment. However, it is minor, please proceed as you wish.

jwuttke:
@gpospelov For the time being, let me go for the convenience of less typing. And for standardizing these option names across our projects. If this turns out to be confusing, or outright dangerous, then we can easily revert to full <project>_ prefix everywhere. Since these options are for developers only no user will complain about changes forth and back.

PR #1042, opened 2020-09-05
Small changes in scalar Fresnel computation + Bugfix

rbeerwerth:
This PR fixes an ugly use-after-free in RTTest.
Furthermore some cleanup of unused variables in the scalar Fresnel kernel is done and the use of indices is changed such that it is a bit easier readable and consistent with the upcoming polarized engine.


rbeerwerth:
ups... push_back creates a copy of course.

jwuttke:
looks good to me

gpospelov (approved):

PR #1043, opened 2020-09-07
Stabilized existing implementation of magnetic Fresnel computation

rbeerwerth:
This PR provides an enhanced numerical stability of the existing implementation of the polarized Fresnel computation. This method is unchanged and as described in the document theory/polarizedSpecular. 
The modification applied in the last commit is described in Section 4 of that document. 
With these changes, the numerical precision is enhanced for samples with more than a few layers as long as all the the layers are not too thick. Due to the rotation of the polarization there is still a significant loss of precision if the layer thickness becomes too large.
The first three commits do minor changes, also in preparation of the new polarized computational kernel, that will supersede this implementation. Even though this implementation will likely not be used in the future it still makes sense to update it, in order to facilitate testing of the new implementation, and especially the Fresnel coefficients.

Additional external testing was performed to verify these changes.

### DIFF
--- a/Core/Basics/Complex.h
+++ b/Core/Basics/Complex.h
@@ -17,7 +17,8 @@
 
 #include <complex>
 
-typedef std::complex<double> complex_t;
+using complex_t = std::complex<double>;
+constexpr complex_t I = complex_t(0.0, 1.0);
 
 //! Returns product I*z, where I is the imaginary unit.
 inline complex_t mul_I(complex_t z)
--- a/Core/Computation/SpecularComputationTerm.h
+++ b/Core/Computation/SpecularComputationTerm.h
@@ -85,7 +85,7 @@ private:
 protected:
     void eval(SpecularSimulationElement& elem, const std::vector<Slice>& slices) const override;
     double intensity(const SpecularSimulationElement& elem,
-                     ISpecularStrategy::single_coeff_t& coeff) const;
+                     const ISpecularStrategy::coefficient_pointer_type& coeff) const;
 };
 
 #endif // BORNAGAIN_CORE_COMPUTATION_SPECULARCOMPUTATIONTERM_H
--- a/Core/Multilayer/SpecularMagneticStrategy.cpp
+++ b/Core/Multilayer/SpecularMagneticStrategy.cpp
@@ -186,7 +186,7 @@ void SpecularMagneticStrategy::propagateBackwardsForwards(
         const size_t i = static_cast<size_t>(index);
         const double t = slices[i].thickness();
         const auto kz = coeff[i].getKz();
-        Eigen::Matrix4cd l = coeff[i].R1 * GetImExponential(kz(0) * t)
+        const Eigen::Matrix4cd l = coeff[i].R1 * GetImExponential(kz(0) * t)
                              + coeff[i].T1 * GetImExponential(-kz(0) * t)
                              + coeff[i].R2 * GetImExponential(kz(1) * t)
                              + coeff[i].T2 * GetImExponential(-kz(1) * t);
@@ -194,9 +194,7 @@ void SpecularMagneticStrategy::propagateBackwardsForwards(
         coeff[i].m_w_min = l * coeff[i + 1].m_w_min;
 
         // rotate and normalize polarization
-        auto r = findNormalizationCoefficients(coeff[i]);
-        auto S = std::get<0>(r);
-        auto norm = std::get<1>(r);
+        const auto [S, norm] = findNormalizationCoefficients(coeff[i]);
 
         SMatrices[i] = S;
         Normalization[i] = norm;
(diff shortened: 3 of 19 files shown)


Core/Computation/SpecularComputationTerm.h
```diff
@@ -85,7 +85,7 @@ class SpecularMatrixTerm : public SpecularComputationTerm
 protected:
     void eval(SpecularSimulationElement& elem, const std::vector<Slice>& slices) const override;
     double intensity(const SpecularSimulationElement& elem,
-                     ISpecularStrategy::single_coeff_t& coeff) const;
+                     const ISpecularStrategy::coefficient_pointer_type& coeff) const;
```
gpospelov, line 88: `pointer&` looks a bit embarrassing for me. I wouldn't encode type inside the definition, this would simplify switching to something different later on (std::variant). Are you using `coefficient_type` anywhere else? One could use `using coefficient_type = std::unique_ptr<const ILayerRTCoefficients>;`. This all of course is minor.

Core/RT/MatrixRTCoefficients_v2.h
```diff
@@ -27,6 +27,7 @@ class BA_CORE_API_ MatrixRTCoefficients_v2 : public ILayerRTCoefficients
 {
 public:
     friend class SpecularMagneticStrategy;
+    friend class SpecularMagnetic_v3ConsistencyTest_ScalarMagneticAmplitudes_Test;
```
gpospelov, line 30: I recon, it is temporary solution?

gpospelov (approved):

Core/Computation/SpecularComputationTerm.h
rbeerwerth: hmm, not entirely sure how you mean this. It would still be an aliases smart pointer, right? Maybe we should avoid these aliases anyhow a bit more...

Core/RT/MatrixRTCoefficients_v2.h
rbeerwerth: yes and now... it was a leftover, i removed it. thanks!

Core/Multilayer/SpecularMagneticStrategy.cpp
```diff
@@ -191,10 +192,54 @@ void SpecularMagneticStrategy::propagateBackwards(std::vector<MatrixRTCoefficien
                              + coeff[i].T2 * GetImExponential(-kz(1) * t);
         coeff[i].m_w_plus = l * coeff[i + 1].m_w_plus;
         coeff[i].m_w_min = l * coeff[i + 1].m_w_min;
+
+        // rotate and normalize polarization
+        auto r = findNormalizationCoefficients(coeff[i]);
+        auto S = std::get<0>(r);
```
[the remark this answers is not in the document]
jwuttke: I'd prefer `r.first` over `get<0>`, as it lets me at least guess that `r` is a `pair`. But why not go for C++17 structured binding: Instead of lines 197-199, just `const auto [SMatrix, norm] = findNormalizationCoefficients(coeff[i]);`

Core/Multilayer/SpecularMagneticStrategy.cpp
```diff
@@ -205,25 +250,11 @@ SpecularMagneticStrategy::findNormalizationCoefficients(const MatrixRTCoefficien
 
     Eigen::Matrix2cd result;
     result << S(1, 1), -S(0, 1), -S(1, 0), S(0, 0);
-    result /= S(0, 0) * S(1, 1) - S(1, 0) * S(0, 1);
-
-    return result;
-}
+    auto d1 = S(1, 1) - S(0, 1);
```
jwuttke, line 253: explicit type, please (double or complex_t ?). And `const` whereever possible.

Core/Multilayer/SpecularScalarStrategy.h
```diff
@@ -34,6 +34,10 @@ class Slice;
 class BA_CORE_API_ SpecularScalarStrategy : public ISpecularStrategy
 {
 public:
+    using coefficient_type         = ScalarRTCoefficients;
```
jwuttke, line 37: What's the advantage of this and the following alias? It's hardly shorter than the right hand side.

Core/RT/ILayerRTCoefficients.h
```diff
@@ -60,6 +60,11 @@ class BA_CORE_API_ ILayerRTCoefficients
         throw Exceptions::NotImplementedException("ILayerRTCoefficients::"
                                                   "getScalarKz(): coefficients are not scalar.");
     }
+
+    virtual Eigen::Matrix2cd getReflectionMatrix() const
```
jwuttke, line 64: Can this be moved to the parent class? Always raise NotImplemented unless defined in child class.

Core/Basics/Complex.h
```diff
@@ -17,7 +17,8 @@
 
 #include <complex>
 
-typedef std::complex<double> complex_t;
+using complex_t = std::complex<double>;
+constexpr complex_t I = complex_t(0.0, 1.0);
```
jwuttke, line 21: Not sure about this. Discounting the cases where `exp_I` can be used, I feel there are not enough explicit uses of `I` in the code to justify this ultrashort global variable.

Core/Multilayer/SpecularMagneticStrategy.h
```diff
@@ -33,6 +33,10 @@ class Slice;
 class BA_CORE_API_ SpecularMagneticStrategy : public ISpecularStrategy
 {
 public:
+    using coefficient_type = MatrixRTCoefficients_v2;
```
jwuttke, line 36: same comment as below (SpecularScalarStrategy.h): is there strong enough an advantage to justify two extra aliases?

jwuttke (changes requested):

Core/Multilayer/SpecularMagneticStrategy.h
rbeerwerth: yep, i needed it in external tests

PR #1052, opened 2020-09-23
Correctly handle kz = 0 and kz -> 0 for polarized computations

rbeerwerth:
This PR corrects the treatment of the case kz = 0 and its limit to be consistent with the scalar implementation.
The tests are corrected in accordance and a proper consistency check between the scalar and polarized computation is performed.
The case k = 0 is not handled correctly in the previous implementation, and is disabled now.


jwuttke:
Looks good to me, though of course I cannot verify the math.

Since I did hit the "rebase" button, you have to update your local branch from here. Then please run `clang-format -i` over the modified source files.

No need to wait for Gennady's return, please merge.

PR #1102, opened 2020-11-14
revise class Lattice

jwuttke:
rm parameterization of lattice orientation by Miller indices (unused and too complicated), rm unused functions; merge functions; give fct arguments more expressive names; avoid confusing a1,a2,a3 vs ma,mb,mc; move static constructing fcts to separate file & namespace.


jwuttke:
to be done along with next one

PR #1115, opened 2020-11-26
normalize function get_sample() in export function and in examples

jwuttke:


jwuttke:
needs more consideration

PR #1122, opened 2020-11-29
Feature 2115 replace boost filesystem

mpuchner:
Implementation of feature 2115: replace boost::filesystem by std::filesystem

I also added a bunch of new unit tests which revealed a few shortcomings in the existing codebase. These have been fixed along with the migration to std::filesystem.

Dependency from boost::filesystem removed from CMakeLists.txt


mpuchner:
There is another reference to boost's filesystem in a comment in "PackDebian.cmake". Maybe this can be removed? Was not sure about it. Perhaps it would be good if the relevant change would be added to this branch as well to have all migration in one place?

jwuttke:
> There is another reference to boost's filesystem in a comment in "PackDebian.cmake". Maybe this can be removed? Was not sure about it. Perhaps it would be good if the relevant change would be added to this branch as well to have all migration in one place?

Yes, absolutely. Please make that little change, then merge this PR.

PR #1130, opened 2020-12-03
Update FindBornAgain.cmake

rbeerwerth:
This updates the include FindBornAgain.cmake file to support the split of core into several shared objects.
Then the cpp example is usable again, and it works to use BornAgain as a library from external C++ programs.

### DIFF
--- a/Examples/cpp/modules/FindBornAgain.cmake
+++ b/Examples/cpp/modules/FindBornAgain.cmake
@@ -17,47 +17,20 @@
 
 
 set(BORNAGAINSYS $ENV{BORNAGAINSYS})
+set(CoreComponents "Base;Param;Sample;Device;Core")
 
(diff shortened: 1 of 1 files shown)


jwuttke:
Thanks for this important correction.

There is quite some repetition in the code. Wouldn't it be appropriate to use a loop
```
foreach(lib ${CoreComponents})
    find_library (BORNAGAIN_${lib}  _libBornAgain${lib}.so PATHS  .. HINTS ..)
    append(BORNAGAIN_LIBRARIES ${BORNAGAIN_${lib}})
endforeach()
```
with `CoreComponents` as defined in top-level CMakeLists.

rbeerwerth:
yes, that seems to be a possiblity. Thx for pointing the existence of CoreComponents out.

PR #1135, opened 2020-12-05
fix "Crash on saving to deleted project"

mpuchner:
Please read the commit comment for details


GUI/coregui/utils/GUIHelpers.cpp
```diff
@@ -188,11 +188,12 @@ QString baseName(const QString& fileName) {
...
-            throw GUIHelpers::Error("GUIHelpers::createSubdir() -> Error. Can't create '"
-                                    + subdirName + "' in parent directory '" + parentName + "'.");
-    }
+    if (projectDir.exists())
+        if (!projectDir.exists(subdirName))
+            if (!projectDir.mkdir(subdirName))
+                throw GUIHelpers::Error("GUIHelpers::createSubdir() -> Error. Can't create '"
+                                        + subdirName + "' in parent directory '" + parentName
+                                        + "'.");
 }
```
jwuttke, line 197: We should use this opportunity to further improve the legacy code.  The actual action (mkdir) should not be encapsulated by tests for anomalous conditions. Rather somewhat like
```
ASSERT(projectDir.exists());
if (!projectDir.exists(subdirName))
    throw ...
bool success = projectDir.mkdir(subdirName));
if (!success)
     throw ...
```
Choose throw vs ASSERT depending on whether or not the condition can be triggered by bad user input.

jwuttke (changes requested):

GUI/coregui/utils/GUIHelpers.cpp
mpuchner: Reworked the dependencies so this code became obsolete

jwuttke: even better, thank you

jwuttke (approved):

PR #1152, opened 2020-12-15
Add Q-Offset to QSpecScan

rbeerwerth:
This PR adds API support for an offset in Q to QSpecScan. This solves http://apps.jcns.fz-juelich.de/redmine/issues/2496 and eliminates the necessity to work around this missing feature as described in https://www.bornagainproject.org/documentation/examples/fitting/extended/reflectometry-pt-layer/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,7 @@ BornAgain-1.18.99, ongoing development
     * Removed some interference functions constructors
     * Remove old R&T computations from API (now in Code/Legacy and Tests)
     * Python plot API entirely keyword based
+    * Add support for Q-Offset in QSpecScan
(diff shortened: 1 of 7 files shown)


CHANGELOG
```diff
@@ -8,6 +8,8 @@ BornAgain-1.18.99, ongoing development
     * Removed some interference functions constructors
     * Remove old R&T computations from API (now in Code/Legacy and Tests)
     * Python plot API entirely keyword based
+  > Reflectometry changes:
```
jwuttke, line 11: No separate section for reflectometry please. We strive to support SAS, GISAS, off-specular, specular through one and the same API.

jwuttke (changes requested):

PR #1155, opened 2020-12-16
Reduce manhattan deps

mpuchner:
Remove any usage of "Manhattan"

The library itself is still present. If everythings seems OK I will remove it in a next PR.

Because of usage of Qt-native concepts instead of Manhattan, there are a few changes visible on the UI. 

Changed handling of dockwidgets also allowed reducing code complexity in this area.

UI Changes:
* New menu "view" to select the dock widgets
* Dockwidgets now always have a title bar
* Moved actions from the bottom-right-combos to the view menus
* Context menus no longer contain the dock widget actions

grafik

grafik


GUI/coregui/mainwindow/mainwindow.cpp
```diff
@@ -338,6 +338,13 @@ void MainWindow::updateViewSelectionButtonsGeometry() const {
         b->setFixedSize(buttonExtent, buttonExtent);
         b->setIconSize({iconExtent, iconExtent});
     }
+    // set fixed width in filler and progress bar
+    auto filler = m_viewSelectionButtonsLayout->itemAt(m_viewSelectionButtons->buttons().size());
+    if (filler != nullptr)
+        if (auto fillerBtn = dynamic_cast<QToolButton*>(filler->widget()); fillerBtn != nullptr)
```
gpospelov, line 354: I would vote for simple `if(filler)` without explicit comparison with `nullptr`. I think this is what clang-tidy will recommend too.

GUI/coregui/Views/CommonWidgets/DocksController.cpp
```diff
@@ -18,23 +18,68 @@
...
+    QString res = text;
+    for (int index = res.indexOf('&'); index != -1; index = res.indexOf('&', index + 1))
+        res.remove(index, 1);
+    return res;
+}
+
+} // namespace
+
+DocksController::DocksController(QMainWindow* mainWindow)
+    : QObject(mainWindow), m_handleDockVisibilityChanges(true), m_mainWindow(mainWindow) {
```
gpospelov, line 42: The order of initialization doesn't coincide with the order of declaration

GUI/coregui/Views/CommonWidgets/DocksController.cpp
```diff
@@ -18,23 +18,68 @@
...
+#include <QSettings>
 #include <QTimer>
-#include <qt-manhattan-style/fancymainwindow.h>
 
-DocksController::DocksController(Manhattan::FancyMainWindow* mainWindow)
-    : QObject(mainWindow), m_mainWindow(mainWindow) {
+namespace {
+static const char dockWidgetActiveState[] = "DockWidgetActiveState";
+static const char StateKey[] = "State";
+static const int settingsVersion = 2;
```
gpospelov, line 30: `static` is redundant in the context of the anonymous namespace. It would be better to remove it to avoid noise.

GUI/coregui/Views/CommonWidgets/DocksController.cpp
```diff
@@ -18,23 +18,68 @@
...
+
+QString stripAccelerator(const QString& text) {
+    QString res = text;
+    for (int index = res.indexOf('&'); index != -1; index = res.indexOf('&', index + 1))
+        res.remove(index, 1);
+    return res;
+}
```
gpospelov, line 37: (Arguably better alternative would be `str.erase(std::remove(str.begin(), str.end(), '&'), str.end());`

jwuttke (approved):
Looks good to me. Why not remove qt-manhattan from ThirdParty with this PR?

gpospelov (approved):
Things looks good on my laptop (ArchLinux+Plasma+2560x2440+scale125%).

I will miss icons at the bottom bar, but I think it was the right move to get rid of them. I have added a few minor comments. You can resolve them now, or address them later if you want. Feel free to merge.

GUI/coregui/mainwindow/mainwindow.cpp
mpuchner: will change it

GUI/coregui/Views/CommonWidgets/DocksController.cpp
```diff
@@ -43,9 +88,8 @@ void DocksController::addWidget(int id, QWidget* widget, Qt::DockWidgetArea area
 }
 
 void DocksController::onResetLayout() {
-    m_mainWindow->setTrackingEnabled(false);
-    QList<QDockWidget*> dockWidgetList = m_mainWindow->dockWidgets();
-    for (QDockWidget* dockWidget : dockWidgetList) {
+    setTrackingEnabled(false);
+    for (QDockWidget* dockWidget : dockWidgets()) {
```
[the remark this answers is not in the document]
mpuchner: I agree! will change it.

GUI/coregui/Views/SampleView.h
```diff
@@ -18,40 +18,42 @@
...
+    DocksController* m_docks = nullptr;
+    SampleToolBar* m_toolBar = nullptr;
 
-    SampleToolBar* m_toolBar;
+    SampleDesigner* m_sampleDesigner = nullptr;
+    SampleWidgetBox* m_widgetBox = nullptr;
+    SampleTreeWidget* m_treeWidget = nullptr;
+    SamplePropertyWidget* m_propertyWidget = nullptr;
+    ScriptPanel* m_scriptPanel = nullptr;
+    RealSpacePanel* m_realSpacePanel = nullptr;
```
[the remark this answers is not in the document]
mpuchner: Well here I wouldn't like it so much - no risk of implicit conversion, less characters, better readable. Why would you prefer the braces?

GUI/coregui/Views/CommonWidgets/DocksController.cpp
mpuchner: will change it

GUI/coregui/Views/CommonWidgets/DocksController.cpp
mpuchner: was a copy&paste from the original - will change it

GUI/coregui/Views/SampleView.h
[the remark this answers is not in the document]
gpospelov: It's up to you.  After watching [Josuttis at CPPCon](https://www.youtube.com/watch?v=7DTlWPgX6zs) I decided to stick to this style. I don't think we have a policy in a group. So I had to write "please consider using braced initializers".

GUI/coregui/Views/CommonWidgets/DocksController.cpp
mpuchner: No, the original does not remove an 'accelerated &':
* &A => A
* &&A => &A

The suggested code would deliver only 'A' in both cases

mpuchner:
> 
> 
> Looks good to me. Why not remove qt-manhattan from ThirdParty with this PR?

Will do so...

PR #1164, opened 2020-12-30
polyhedral form factors: rm numeric trick; improve messages from failed tests

jwuttke:


jwuttke:
The numeric trick was in a sum_1^n a_i  b_i, and consisted in replacing a_n by -sum_1^{n-1} a_i. Upon closer inspection, I found no justification for, and no advantage of, using this trick.

jwuttke:
At this occasion, the q battery is reorganized, and all numeric tests are labelled XL.

PR #1172, opened 2021-01-13
 Replace any SIGNAL() & SLOT() by the non-macro-versions

gpospelov:
- QCustomPlot left intact
- Views/widgetbox left intact (legacy code borrowed from Qt-creator)


mpuchner (approved):
Looks good to me, thanks!

PR #1188, opened 2021-01-21
Decrease static string usage

gpospelov:
Following changes toward better type safety.

+ Switch to item<T> access method instead of getItem(P_SOMETHING)
+ Switch to SessionModel::insertItem<T> instead of insertNewItem(QString)


mpuchner (approved):
Looks good (of course I did not check every line...)

gpospelov:
Tx!

PR #1189, opened 2021-01-22
Insert new item review

gpospelov:
- Replace most of remaining methods `insertNewItem` toward more type safe `insertItem<>`
- Change signature of `insertItem(QModelIndex parent)` to insertItem(SessionItem* parent)`

This is done to make an interface more uniform with other methods of `SessionModel` (`copyItem`, `moveItem`, etc), and avoid repeating code like

```
model.insertItem<Child>(parent->index())
or
model.insertItem<Child>(model.indexOfItem(parent))
```

in the favor of

```
model.insertItem<Child>(parent)
```

### DIFF
--- a/GUI/coregui/Views/JobWidgets/ProjectionsEditorCanvas.cpp
+++ b/GUI/coregui/Views/JobWidgets/ProjectionsEditorCanvas.cpp
@@ -90,9 +90,11 @@ void ProjectionsEditorCanvas::onEnteringColorMap()
     m_block_update = true;
 
     if (m_currentActivity == MaskEditorFlags::HORIZONTAL_LINE_MODE)
-        m_liveProjection = m_model->insertItem<HorizontalLineItem>(m_containerIndex);
+        m_liveProjection =
+            m_model->insertItem<HorizontalLineItem>(m_model->itemForIndex(m_containerIndex));
     else if (m_currentActivity == MaskEditorFlags::VERTICAL_LINE_MODE)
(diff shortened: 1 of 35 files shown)


GUI/coregui/Views/JobWidgets/ProjectionsEditorCanvas.cpp
```diff
@@ -90,9 +90,11 @@ void ProjectionsEditorCanvas::onEnteringColorMap()
     m_block_update = true;
 
     if (m_currentActivity == MaskEditorFlags::HORIZONTAL_LINE_MODE)
-        m_liveProjection = m_model->insertItem<HorizontalLineItem>(m_containerIndex);
+        m_liveProjection =
+            m_model->insertItem<HorizontalLineItem>(m_model->itemForIndex(m_containerIndex));
```
mpuchner, line 94: This seems nicer in the old version. Same in the following spots, however, I did not add a review comment to all of them.
Suggestion: Use the old method, and add a comment to the method's docu that it is deprecated (index usage discouraged). Maybe also add a // #migration tag to mark it for refactoring

mpuchner (commented):
I like it. Please think of the slight change I proposed.

GUI/coregui/Views/JobWidgets/ProjectionsEditorCanvas.cpp
gpospelov: Implemented in the next commit.

mpuchner (approved):
Thanks

PR #1192, opened 2021-01-26
Decrease get item usage

gpospelov:
- Few getter/setters for axes related properties.
- Replace getItem with type-safe item<T> (where it was easy to do).
- Number of `getItem()` usages decreased from 385 to 292.

### DIFF
--- a/GUI/coregui/Models/SpecularDataItem.cpp
+++ b/GUI/coregui/Models/SpecularDataItem.cpp
@@ -56,17 +56,17 @@ void SpecularDataItem::setOutputData(OutputData<double>* data)
 
 int SpecularDataItem::getNbins() const
 {
-    return xAxisItem()->getItemValue(BasicAxisItem::P_NBINS).toInt();
+    return xAxisItem()->binCount();
 }
 
 double SpecularDataItem::getLowerX() const
 {
-    return getItem(P_XAXIS)->getItemValue(BasicAxisItem::P_MIN_DEG).toDouble();
+    return xAxisItem()->lowerBound();
 }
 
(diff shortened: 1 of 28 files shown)


GUI/coregui/Models/SpecularDataItem.cpp
```diff
@@ -233,9 +233,12 @@ BasicAxisItem* SpecularDataItem::xAxisItem()
...
 {
-    auto result = dynamic_cast<const AmplitudeAxisItem*>(getItem(P_YAXIS));
-    ASSERT(result);
-    return result;
+    return item<AmplitudeAxisItem>(P_YAXIS);
+}
+
+AmplitudeAxisItem* SpecularDataItem::yAxisItem()
+{
+    return const_cast<AmplitudeAxisItem*>(static_cast<const SpecularDataItem*>(this)->yAxisItem());
```
mpuchner, line 241: What about ```return item<AmplitudeAxisItem>(P_YAXIS);```? If you agree on it, then please also for ```xAxisItem()```

mpuchner (commented):
Looks a lot nicer than before :-) Thanks!

mpuchner (approved):

GUI/coregui/Models/SpecularDataItem.cpp
gpospelov: The original recipe came from Scott Meyer's `Effective C++`. It's not elegant, but dogmatic. Nevertheless, following the request and replacing it with a simpler version (though introducing code duplication). Would be nice to learn one day the recipe which would be elegant, and not containing code duplication.

PR #1193, opened 2021-01-27
Add single property items (i.e. VectorItem) via type safe mechanism

gpospelov:
Provides the possibility to add a single property item via the type-safe addProperty<> method.
The final task is to replace addGroupProperty completely with a type-safe mechanism.

- [x] Number off addGroupProperty usages 55 -> 27.
- [x] Number off getItem usages 292 -> 257.


mpuchner (approved):
Looks good. Thanks for adding #migration tags!

gpospelov:
Thanks!

Issue #1200, opened 2021-02-02
Spin-flip reflectivity from non-magnetic sample


aglavic:
When running reflectivity simulations on non-magnetic samples, spin-flip channels have the same intensity as non spin-flip from the nuclear structure.

### Expected behavior
spin-flip intensity should be zero if polarization efficiency is 100%

### Actual behavior
spin-flip = non spin-flip

rbeerwerth:
Thanks for reporting this issue!
This happens when setScan is called after setting polarization and analyzer properties. As a workaround make sure to call setScan immediately after creating the SpecularSimiulation object.

PR #1217, opened 2021-02-22
Correct handling of polarization

rbeerwerth:
This PR corrects the handling of polarization.
* Fixes the problem reported #1200: If an analyzer is set now the polarized engine is utilized even when the sample itself is non-magnetic. This ensures that the spin-flip intensity vanishes, as expected.
* This applies to reflectometry as well as scattering on particles
* Also problems when changing the of some initialization of the beam parameters is fixed.

This needs independent validation.


Core/Simulation/ISimulation.cpp
```diff
@@ -186,6 +186,7 @@ void ISimulation::setAnalyzerProperties(const kvector_t direction, double effici
                                         double total_transmission)
 {
     detector().setAnalyzerProperties(direction, efficiency, total_transmission);
+    m_options.setIsPolarized(true);
```
jwuttke, line 189: What if a user calls simulation.detector().setAnalyzerProperties instead of simulation.setAnalyzerProperties?
Generally, we are converting towards indirect calls; we should rather remove ISimulation::setAnalyzerProperties for good.
So need a different mechanism to replace the proposed ISimulation::m_options.

Sample/RT/SimulationOptions.h
```diff
@@ -62,10 +62,15 @@ class SimulationOptions {
...
 
+    void setIsPolarized(bool is_polarized) { m_is_polarized = is_polarized; }
+
+    bool isPolarized() const { return m_is_polarized; }
+
 private:
     bool m_mc_integration;
     bool m_include_specular;
     bool m_use_avg_materials;
+    bool m_is_polarized;
```
jwuttke, line 73: As far as I see, the other member variables are user choices whereas m_is_polarized caches information deduced from other settings.

rbeerwerth: Not sure what you mean by user choices. But `simulation.getOptions().setUseAvgMaterials(True)` sets `m_use_avg_materials`.

jwuttke (changes requested):
Fix of #1200 is highly appreciated.
However, let's clearly distinguish user-chosen options and cached derived information; and keep analyzer properties out of Simulation classes.

Sample/RT/SimulationOptions.h
jwuttke: Maybe use_avg_materials should also be simplified?

rbeerwerth:
closing.
New clean PR in preparation

PR #1219, opened 2021-02-26
Remove dependency from MVVM lib

mpuchner:
MVVM dependency with current MAC pipeline causes runtime error on MAC (missing library when installing with one of the artifacts). This may be the case as well on Windows/Linux, but I didn't test it.

Since so far only one minor call is used from MVVM, this dependency has been removed.

Seems important for upcoming release 1.19.0


jwuttke (approved):
of course I fully agree

PR #1223, opened 2021-03-03
Correct handling of polarization + update unit tests

rbeerwerth:
This PR corrects the handling of polarization.
* Fixes the problem reported #1200: If an analyzer is set now the polarized engine is utilized even when the sample itself is non-magnetic. This ensures that the spin-flip intensity vanishes, as expected.
* This applies to reflectometry as well as scattering on particles
* Also problems when changing the order of some initialization of the beam parameters is fixed.
* `setAnalyzerProperties` removed from ISimulation and SpecularSimulation
* On the way some unit tests were extended to cover polarization and the related properties
* On the way some other untested properties were added primarily to clone tests


jwuttke (approved):
Danke

